PR #16, opened 2026-03-25
Add collapsed transcript sections for tool and file-change activity

jeremytondo:
## Summary
- Group transcript turn items into ordered `Tools` and `File Changes` sections on the Swift side without changing bridge payloads
- Render grouped activity as collapsible section cards while keeping assistant and reasoning items in their existing transcript positions
- Auto-expand running activity, default completed sections to collapsed, and add accessibility identifiers for grouped sections and toggles
- Add unit and UI coverage for grouping behavior, aggregate section status, disclosure defaults, and transcript ordering

## Testing
- `BuildProject` via Xcode MCP
- `RunSomeTests`: `ThreadSessionTests/*`
- `RunSomeTests`: `WorkspaceBridgeRuntimeTests/structuredTurnEventsPopulateRichSessionState()`
- `RunSomeTests`: updated phase 2 UI tests
- `xcodebuild test -project AtelierCode.xcodeproj -scheme AtelierCode -destination 'platform=macOS'`

### DIFF
--- a/AtelierCode/AppBootstrap.swift
+++ b/AtelierCode/AppBootstrap.swift
@@ -145,11 +145,11 @@ private final class UITestWorkspaceRuntime: WorkspaceConversationRuntime {
                 DiffFileChange(id: "phase2-session", path: "AtelierCode/ThreadSession.swift", additions: 18, deletions: 4)
             ]
 
-            try await Task.sleep(nanoseconds: 30_000_000)
+            try await Task.sleep(nanoseconds: 900_000_000)
             session.appendAssistantTextDelta("I grouped the current turn details under the transcript.")
             session.appendThinkingDelta("Inspecting activity, approvals, plan state, and diff summaries for the active turn.")
             session.startActivity(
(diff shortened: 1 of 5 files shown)


jeremytondo:
Couple pieces of UI feedback.

1. I think there is a lot of duplicate information in each tool call card. For example, it shows the exit code more than once and tne actual tool call more than once. I don't think we need that kind of duplication. Let's tighten it up a bit.

2. When one tool call in a group fails, the whole thing shows as failed. I think we should find a better way to display this that shows how may failed and how many succeeded if both are true in the group.
Screenshot 2026-03-24 at 5 11 14 PM

3. I don't think we need to include the completed pill on the agent responses.

4. I think we need some kind of indication that the agent is working. 

I like the three dots for general agent waiting time to show inline 
Screenshot 2026-03-24 at 5 13 48 PM

I like the spinners for tool call cards and card group waiting time.  
Screenshot 2026-03-24 at 5 13 21 PM

PR #48, opened 2026-04-08
App Server Phase 1

jeremytondo:
## Summary
- add a new `AppServer/` Bun + TypeScript module for Phase 1 of the App Server replacement work
- document the App Server implementation foundations, module boundaries, validation model, persistence strategy, and testing strategy in `Docs/app-server-design.md`
- implement the Phase 1 JSON-RPC-shaped WebSocket contract for `initialize`, `workspace/open`, `thread/start`, and `turn/start`
- add deterministic in-memory workspace/thread/turn/item state plus a fake agent adapter so the protocol can be exercised without a live Codex runtime
- add broad schema, protocol, domain, service, and websocket harness coverage around the new server foundation

## What Changed
- scaffolded `AppServer/` as a standalone Bun package with Biome, TypeScript, Bun lockfile, and server/check/typecheck/healthcheck scripts
- added the Phase 1 server boundaries under `transport/`, `protocol/`, `schema/`, `domain/`, `server/`, `store/`, `agent-adapters/`, and `test-support/`
- implemented envelope parsing, request dispatch, response/error serialization, and protocol notification emission for the initial method set
- added schema validation for `initialize`, `workspace/open`, `thread/start`, and `turn/start`, plus protocol-level validation for malformed JSON, invalid envelopes, unsupported methods, and execution-rule failures
- added in-memory store and session-state handling that enforce initialize-before-use, workspace ownership, stable IDs, one active turn per thread, and pending approval scoping
- implemented a deterministic fake agent adapter and runtime turn runner that emit Codex-shaped `thread/started`, `turn/started`, `item/started`, `item/agentMessage/delta`, `item/completed`, and `turn/completed` notifications
- added websocket harness helpers that spawn the server process, connect over `ws`, and assert healthcheck behavior, request/response flow, notification ordering, and invalid-sequencing errors
- updated `AGENTS.md` with App Server-specific implementation standards and completion checks

## Testing
- `cd AppServer && bun run check`
- `cd AppServer && bun run typecheck`
- `cd AppServer && bun test`


jeremytondo (commented):
I ran `bun run check`, `bun run typecheck`, and `bun test` in `AppServer/`; all passed.

I still found four foundation-level issues that are worth addressing before Phase 2 and later lifecycle/persistence phases build on this module:

1. **[P1] Unhandled turn follow-up failures escape the protocol boundary**
   `outcome.followUp()` is launched in a detached microtask with no error handling in `AppServer/src/transport/websocket-server.ts`. Today that path runs the `turn/started` notification and then `RuntimeTurnRunner.run()`, so any adapter/runtime/store exception after the initial response bypasses JSON-RPC error mapping, never marks the turn failed, and can surface as an unhandled rejection once the real Codex adapter starts failing in normal ways.

2. **[P1] Protocol thread/turn types are phase-1 stubs instead of canonical contract types**
   `ProtocolTurn.items` and `ProtocolThread.turns` are typed as literal empty tuples in `AppServer/src/protocol/types.ts`, and `ProtocolThread` omits several fields that already exist in the vendored Codex `Thread` contract. That matches the current happy path, but it hardcodes the Phase 1 subset into the shared protocol model, so `thread/read`/`thread/resume`/real-adapter work will require widening the wire types instead of simply filling in already-canonical contracts.

3. **[P2] Outbound notifications are not validated against per-method schemas**
   The design says outbound notifications should be validated against canonical schemas, but the guard in `AppServer/src/protocol/message-assertions.ts` only checks that the method name is known and that `params` is an object. A future adapter can therefore emit a malformed `thread/started`, `item/completed`, or `turn/completed` payload without any boundary validation, which weakens the contract right before Phase 2 integration work.

4. **[P2] Workspace identity depends on the raw input path string**
   `workspace/open` in `AppServer/src/server/app-server-service.ts` only reuses an existing workspace when `params.path` matches exactly. Because the path abstraction only checks `isDirectory()` and never canonicalizes, the same directory can be opened through different spellings such as relative paths, symlinks, or trailing slash variants and get different workspace IDs, which will make persistence and restart reattachment much harder to keep correct.

The broad direction looks strong, but these are the places I’d tighten before treating this as the long-term foundation for the next App Server phases.

PR #87, opened 2026-07-11
Fix terminal reconnect and flow control

jeremytondo:
## What changed

- Automatically reconnect transient terminal attach failures with capped exponential backoff (nominal 500 ms to 8 s, ±20% jitter) and a 10-attempt budget per failure streak, while leaving ended sessions and explicit disconnects stopped. A dedicated Reconnecting phase keeps the status banner steady across backoff cycles; once the budget is spent the banner offers a manual Reconnect, and wake/network recovery restores the budget automatically.
- Re-attach on macOS wake and recovered network paths, shorten the WebSocket ping interval to 10 seconds, and ignore late events from superseded connections.
- Rebuild the Ghostty in-memory backend only after a replacement transport opens, immediately before the resize that triggers zmx snapshot replay. Failed reconnect attempts keep the existing screen and scrollback visible.
- Replace the client’s silent 1 MiB paste truncation with an 8 MiB bounded queue that applies synchronous backpressure, preserves whole-write ordering, and emits 256 KiB WebSocket frames. Empty writes return immediately instead of parking behind a full queue.
- Extend the server’s PTY-to-WebSocket dead-write bound from 10 seconds to 2 minutes so temporary stalls use normal TCP/PTY backpressure instead of forcing a reconnect.
- Add a server-side ping/pong liveness probe (30-second interval, 2-minute pong bound) so a client that vanished without closing TCP cannot hold its attach bridge and zmx child indefinitely on an idle terminal.

## Why

The terminal session survived on the server, but the client did not recover after sleep or network loss. Reconnects then replayed zmx’s complete terminal snapshot into an already-populated Ghostty surface, duplicating screen and scrollback content. Separately, outbound client input silently dropped when its queue crossed 1 MiB, and the server converted recoverable write stalls into disconnects after 10 seconds.

The reconnect reset timing follows zmx v0.6.0’s attach flow: terminal state is serialized on re-attach before resize and includes scrollback plus the visible screen.

The retry budget bounds churn on permanent failures (revoked token, deleted server) without sacrificing unattended recovery: wake and network-path transitions reset the budget, so any environment that heals later reconnects on its own.

## Validation

- `TMPDIR=/tmp mise run check`
  - server formatting, vet, and Go tests (including dead-client reaping and healthy-client survival under fast pings)
  - web type check and 8 tests
  - ATCKit: 46 tests
  - macOS: 125 tests
- Focused terminal suites: 11 tests covering retry policy, retry-budget exhaustion and restoration, reconnecting-phase stability, terminal stop states, stale-event gating, wake/path recovery, delayed surface replacement, large-paste integrity, ordered backpressure, empty-write fast path, and teardown wakeup.
- Independent full-diff subagent review completed; all actionable findings addressed in the follow-up polish commit.

## Issues

- Fixes [DEV-25](https://linear.app/elevenideas/issue/DEV-25/terminal-sessions-dont-auto-reconnect-after-sleepnetwork-drop)
- Fixes [DEV-26](https://linear.app/elevenideas/issue/DEV-26/reconnect-visually-corrupts-terminal-screen-duplicated-snapshot)
- Fixes [DEV-27](https://linear.app/elevenideas/issue/DEV-27/reconsider-the-10s-attach-write-timeout-it-drops-recoverable-stalls)
- Fixes [DEV-29](https://linear.app/elevenideas/issue/DEV-29/large-pastes-silently-truncate-at-1-mib-instead-of-applying)

### DIFF
--- a/macos/ATC/AppModel.swift
+++ b/macos/ATC/AppModel.swift
@@ -206,8 +206,4 @@ final class AppModel {
             selection = nil
         }
     }
-
-    deinit {
-        terminalRecoveryMonitor.stop()
-    }
 }
--- a/macos/ATC/AppModel.swift
+++ b/macos/ATC/AppModel.swift
@@ -33,10 +33,14 @@ final class AppModel {
     private(set) var terminals: [SessionRef: TerminalSessionController] = [:]
 
     private let clientFactory: (ConnectionRecord) -> any ATCClient
+    private let terminalControllerFactory: (String, any ATCClient) -> TerminalSessionController
+    private let terminalRecoveryMonitor: TerminalRecoveryMonitor
 
     init(
         connections: ConnectionsStore? = nil,
-        clientFactory: ((ConnectionRecord) -> any ATCClient)? = nil
+        clientFactory: ((ConnectionRecord) -> any ATCClient)? = nil,
+        terminalControllerFactory: ((String, any ATCClient) -> TerminalSessionController)? = nil,
+        terminalRecoveryMonitor: TerminalRecoveryMonitor? = nil
     ) {
         self.connections = connections ?? ConnectionsStore()
         self.clientFactory = clientFactory ?? { record in
@@ -45,11 +49,19 @@ final class AppModel {
             let url = URL(string: record.urlString)!
             return HTTPATCClient(server: ATCServer(baseURL: url, token: record.token))
         }
+        self.terminalControllerFactory = terminalControllerFactory ?? { sessionID, client in
+            TerminalSessionController(sessionID: sessionID, client: client)
+        }
+        self.terminalRecoveryMonitor = terminalRecoveryMonitor ?? TerminalRecoveryMonitor()
         for record in self.connections.connections {
             if let runtime = makeRuntime(record) {
                 runtimes.append(runtime)
             }
         }
+        self.terminalRecoveryMonitor.onRecovery = { [weak self] in
+            self?.recoverTerminalsAfterInterruption()
+        }
+        self.terminalRecoveryMonitor.start()
     }
 
     // MARK: - Runtime access
@@ -148,12 +160,21 @@ final class AppModel {
         guard session.attachable,
               terminals[ref] == nil,
               let runtime = runtime(id: connectionID) else { return }
-        terminals[ref] = TerminalSessionController(sessionID: session.id, client: runtime.client)
+        terminals[ref] = terminalControllerFactory(session.id, runtime.client)
     }
 
     func disconnectTerminal(ref: SessionRef) {
         terminals[ref]?.disconnect()
         terminals.removeValue(forKey: ref)
+    }
+
+    /// Wake and path recovery are app-wide signals. A controller decides
+    /// whether it is still expected to be live; ended sessions and explicit
+    /// disconnects therefore remain stopped.
+    func recoverTerminalsAfterInterruption() {
+        for controller in terminals.values {
+            controller.recoverAfterInterruption()
+        }
     }
 
     // MARK: - Private
@@ -184,5 +205,9 @@ final class AppModel {
         if selection?.connectionID == runtime.id {
             selection = nil
         }
+    }
+
+    deinit {
+        terminalRecoveryMonitor.stop()
     }
 }
--- a/macos/ATC/Features/Terminal/TerminalPane.swift
+++ b/macos/ATC/Features/Terminal/TerminalPane.swift
@@ -46,6 +46,11 @@ struct TerminalStatusBanner: View {
             }
         case .connected:
             EmptyView()
+        case .reconnecting:
+            banner {
+                ProgressView().controlSize(.small)
+                Text("Reconnecting…")
+            }
         case .ended(.sessionEnded):
             banner {
                 Image(systemName: "checkmark.circle")
--- a/macos/ATC/Features/Terminal/TerminalRecoveryMonitor.swift
+++ b/macos/ATC/Features/Terminal/TerminalRecoveryMonitor.swift
@@ -12,7 +12,10 @@ final class TerminalRecoveryMonitor {
     private let wakeNotification: Notification.Name
     private let pathMonitor: NWPathMonitor?
     private let pathQueue = DispatchQueue(label: "ElevenIdeas.atc.terminal-path")
-    private var wakeObserver: (any NSObjectProtocol)?
+    // Read from the nonisolated `deinit` to release the observer. Only
+    // mutated on the main actor while the monitor is alive, and deinit runs
+    // strictly after the last reference is gone, so this cannot race.
+    private nonisolated(unsafe) var wakeObserver: (any NSObjectProtocol)?
     private var previousPathWasSatisfied: Bool?
     private var started = false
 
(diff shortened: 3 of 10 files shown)


jeremytondo:
## Code Review

### Overview

The PR does what it says: client-side auto-reconnect with capped exponential backoff (500ms–8s, ±20% jitter), wake/network-path recovery signals, generation-gated event handling, a deferred Ghostty backend swap that avoids duplicated snapshot replay, an 8 MiB blocking outbound queue replacing silent 1 MiB truncation, and a 10s→2min server write timeout. The design is coherent and the hard parts are handled carefully.

### What I verified and found solid

- **No main-thread deadlock via `ConnectionRef`**: `enqueue` fetches the handle under the lock but calls the potentially-blocking `enqueue(data)` *outside* it, so `connectionRef.set(nil)` from the main actor never blocks behind a backpressured producer.
- **Retry-task invariants hold**: every guard-fail path in `scheduleAutomaticReconnect`'s task that leaves `retryTask` non-nil is preceded by `cancelRetry()`, so scheduling can't get permanently wedged.
- **Backend-swap vs. drain-task race is safe**: the drain loop's post-`surfaceIsReady` portion is synchronous on the main actor, so `prepareSurfaceForReplayIfNeeded` can't interleave mid-drain; the `let terminalSession = terminalSession` capture prevents draining into the wrong backend.
- **`OutboundQueue` locking is correct**: `producerLock` serializes whole writes (preserving whole-write ordering), `finish()` broadcasts to wake blocked producers, and the pump's send-failure path calls `outbound.finish()` so a producer can't stay blocked behind a dead pump. `dequeue`'s O(1) amortized compaction is a nice touch.
- **Concurrent attach on the server is fine**: each attach spawns its own `zmx attach` child with its own PTY, so a stale 2-minute bridge doesn't block a reconnecting client.
- The `nonisolated` annotations are consistent with the project's `SWIFT_DEFAULT_ACTOR_ISOLATION = MainActor` setting.

### Findings

**1. Unbounded retry on permanent failures** (design gap, worth deciding deliberately)
`.serverError` and `.transportFailure` retry forever at the 8s cap (`TerminalSessionController.swift:212-214`). A revoked token, a deleted server, or a rejected handshake all surface as `transportFailure`, so the controller churns indefinitely — each attempt allocating a fresh `URLSession`. Related UX: `phase` flips between `.ended(...)` and `.connecting` on every cycle with no `.reconnecting`-style state, so the indicator will flicker. Consider either an attempt cap with a "gave up — click to retry" terminal state, or at least distinguishing handshake-level auth failures (which will never heal) from network drops.

**2. Every wake tears down and replays every terminal** (deliberate, but know the cost)
`recoverAfterInterruption` unconditionally replaces the connection on every `didWakeNotification` — including a 5-second lid close where all sockets are healthy. Each wake costs a full snapshot replay and Ghostty surface rebuild per attached terminal. The comment justifying it (URLSession's stale-socket problem) is real, and the preserve-screen-until-transport-opens design keeps it visually cheap. Acceptable; a ping-first-reconnect-on-failure refinement is available later if wake flashes annoy.

**3. Ghostty producer thread can block on a stalled-but-alive socket** (bounded, acknowledged)
A >8 MiB paste over a connection that stalls without failing blocks the Ghostty write callback in `condition.wait()` until something calls `finish()`. The escape hatches (10s pings, wake/path recovery closing the old connection) bound this in practice. One caveat: the ping-loop comment says "a failed ping tears the connection," but `URLSessionWebSocketTask.sendPing` failure does not reliably fail the pending `receive()` — pre-existing behavior, just don't over-rely on it.

**4. Nonisolated deinits touching MainActor state** (minor)
`AppModel.deinit` calls the MainActor-isolated `terminalRecoveryMonitor.stop()`, and `TerminalRecoveryMonitor.deinit` touches isolated properties. This compiles under Swift 5 mode but would be an error in Swift 6 language mode. Since the monitor's own `deinit` already duplicates the cleanup, `AppModel.deinit` could simply be deleted.

**5. Small nits**
- `OutboundQueue.enqueue` with empty `Data` on a full queue blocks until space frees despite having nothing to write — an early empty-data check would close that edge.
- Attempt-0 delay with low jitter is 400ms, below the PR body's advertised "500 ms" floor. Cosmetic.
- `beginReconnect` increments `connectionGeneration` and then `connect()` increments it again — harmless, but the first increment is redundant.

**6. Server timeout change** (reasonable, one suggestion)
The 2-minute `attachWriteTimeout` is well-reasoned and well-commented. Note the server now leans entirely on client pings plus this timeout for liveness — a genuinely dead client with no PTY output holds its bridge indefinitely (true before too, since the timeout only arms on writes). A server-side `conn.Ping` loop would be a complementary follow-up, not a blocker.

### Tests

Coverage is genuinely good: deterministic retry-policy math, terminal-end states never reviving, stale-event generation gating, backend identity swapped only after transport opens, large-paste integrity under the frame cap, cross-producer ordering under backpressure, and `finish()` waking a blocked producer. The injection seams (`connectionFactory`, `retrySleep`, `jitterUnit`, `terminalControllerFactory`, disabled monitor) are clean and keep tests hermetic.

### Verdict

**Approve.** The concurrency-heavy parts — the ones most likely to hide bugs — check out under tracing. Finding 1 (infinite retry on permanent failures) is the only thing worth addressing before or shortly after merge; everything else is polish or deliberate tradeoffs documented in-code.

---
*Review by Claude Code*

jeremytondo:
## Review follow-up (6773ca8e)

Disposition of the review findings:

**1. Unbounded retry on permanent failures — fixed.** `RetryPolicy` now carries a 10-attempt budget per failure streak (~1 minute of backoff). When it's spent, the controller settles into the existing "Disconnected — Reconnect" banner instead of churning forever. The budget resets on manual Reconnect *and* on wake/network-path recovery, so an environment that heals later still reconnects unattended. The related flicker is also gone: a new `.reconnecting` phase holds steady across backoff cycles (counts as actively attached for indicators) rather than cycling `.ended` → `.connecting`. Handshake-auth discrimination was deliberately skipped — the WS handshake status isn't reliably observable through `URLSessionWebSocketTask`, and the budget bounds that churn anyway.

**2. Wake tears down every terminal — kept as-is** per the review's own assessment; ping-first refinement remains available later.

**3. Producer block on stalled socket — comment corrected.** The misleading "a failed ping tears the connection" comment now states that `sendPing` failure is not a reliable disconnect signal and that the receive loop plus recovery monitor own failure detection. The blocking itself stays bounded by the existing escape hatches, as the review noted.

**4. Nonisolated deinits — fixed.** `AppModel.deinit` deleted (the monitor's own deinit already does the cleanup). `TerminalRecoveryMonitor.wakeObserver` is now an explicit, documented `nonisolated(unsafe)` opt-out so the deinit is Swift-6-clean; the other properties it touches are Sendable `let`s, which are already legal.

**5. Nits — all fixed.** Empty writes return from `OutboundQueue.enqueue` immediately even when the queue is full; the redundant `connectionGeneration` increment in `beginReconnect` is gone; the PR body now says "nominal 500 ms" with the jitter spelled out.

**6. Server liveness — implemented rather than deferred.** `attachSession` now runs a ping/pong probe (30 s interval, pong bound matching the 2-minute write timeout, so recoverable stalls get the same grace). A client that vanishes without closing TCP no longer holds its bridge and zmx child forever on an idle terminal.

New tests: retry-budget exhaustion + restoration, reconnecting-phase stability with a pending backoff, empty-write fast path, Go dead-client reaping, and Go healthy-client survival under fast pings. `TMPDIR=/tmp mise run check` is green across server, web, ATCKit, and macOS (125 tests).

PR #89, opened 2026-07-12
Workspaces Phase 2: macOS Dashboard and Workspace shell

jeremytondo:
Implements [docs/workspaces-phase2-spec.md](docs/workspaces-phase2-spec.md) — the macOS Dashboard and Workspace shell that replace the throwaway Phase 1 bridge. All work lands in `macos/`; no server, CLI, web, or ATCAPI contract changes.

## Navigation and window architecture
- `RootView` replaces `ContentView`: the Workspace shell is mounted once a Workspace is opened and kept mounted for the window's life, with the Dashboard as an opaque cover — terminal surfaces and WebSockets survive Dashboard round-trips without replay.
- New per-window `WindowState` (`route`, sheet routing, command availability) as the seam for a future multi-window pass; `WorkspaceRef` joins `SessionRef`/`ProjectRef`.
- A vanished open Workspace (remote delete, removed Connection) routes the window back to the Dashboard.

## Data layer
- New `WorkspacesStore` (ProjectsStore pattern) with create/rename/archive/unarchive/delete; `ConnectionRuntime.refresh()` now fetches projects, sessions, workspaces, and actions concurrently, so classification and the creation sheets never rely on stale on-demand data.
- Archived filtering moved wholly to the view layer; `AppModel.includeArchived` and store-level filtering are gone.
- `SessionsStore.unarchive/delete` and `ProjectsStore.delete` wrappers (delete removes the row locally on success).
- **LRU attachment budget** (12): attaching past the budget evicts the least-recently-used terminal through the standard disconnect path; the selected session and the open Workspace's sessions are pinned — if pins alone exceed the budget, the budget is exceeded rather than dropping a live surface.
- `SessionKind`: pure client-side agent/terminal classification against the Actions registry, with the display-name fallback rules (unnamed shell → "Terminal", unnamed action session → action label, unresolvable action → raw name).

## UI
- **Dashboard**: Connection sections (reachability dot, Local/remote context, Retry when unreachable), Project cards with Workspace rows (newest-first), Show Archived toggle, empty states, arrow-key + Return-to-open.
- **CreateWorkspaceSheet**: one shared flow for all three contexts (fixed project from a card, preselected-changeable from the shell, context-free from the File menu); creation opens the new Workspace.
- **Workspace shell**: Sessions/Terminals sidebar sections with per-section `+`, archived footer toggle, search, per-Workspace selection restore via UserDefaults; content header gains the Action label plus Unarchive/Delete.
- **New Session / New Terminal sheets** replace `CreateSessionSheet`/`NewSessionContext`: Agent Actions only vs Interactive Shell + general Actions; server-default Environment; inline error + Retry when the Action list fails to load.
- **Lifecycle**: single-click archive/unarchive; every delete confirms with the target named and ends with the ADR 0008 sentence "Files on disk are not touched."; Workspace delete shows local-store session counts; Project delete only with zero Workspaces.
- **Commands scene**: File (New Project, New Workspace ⌘⇧N, New Session ⌘N, New Terminal ⌘T) and View (Toggle Sidebar ⌘B, Refresh ⌘R) menus shaped for the future shortcut registry; New Project is menu-only now.

## Tests
160 tests green (`mise run macos:test`), including new suites: `DashboardGroups`, `SessionKind`, `WorkspacesStore` (stateful client), `Attachment budget`, `Workspace flows` (routing, command availability, selection memory, delete-confirmation copy), plus rewritten hosting smoke tests for the root/Dashboard/shell and all creation sheets.


jeremytondo:
## Code review results

A thorough automated review was run against the spec. Outcome:

**Fixed in the follow-up commit:**
1. **(major)** Removing the open Workspace's Connection left the window stranded on an empty shell: `AppModel.teardown` nils `openWorkspace` before the view reacts, which defeated the `openWorkspace != nil` guard in `RootView`. The route-back is now keyed on the route itself, with a hosted regression test (`removedConnectionRoutesBack`).
2. **(minor)** The "No Projects" empty overlay appeared when a Connection had only archived Projects hidden behind the toggle. The overlay now keys on `DashboardGroups.totalProjectCount` (unfiltered).
3. **(nit)** `attachOrder` accumulated refs that never attached (selection of cold sessions, non-attachable sessions). LRU marking now happens only for attached refs, so the order stays a permutation of `terminals.keys`.
4. **(nit)** The Stop confirmation said "agent process" even for Terminals; now neutral.
5. **(test gap)** Selection restore is now a testable rule (`WorkspaceSelectionMemory.restoredSelection`) with coverage for surviving-session restore, gone-session fallback, and moved-session fallback.

**Deliberately kept as-is:**
- Reachability still derives from projects + sessions only (spec: "reachability derivation is unchanged"); including workspaces would mark servers predating the workspace routes as unreachable. Now documented at the derivation site.
- The Workspace delete confirmation omits the "0 running sessions will be stopped." clause when nothing is running — a readability deviation from the spec's literal string.

163 tests in 20 suites green after the fixes.

PR #121, opened 2026-07-30
ATC-85: Scaffold the App Server package, serve skeleton, and CI

jeremytondo:
Implements [ATC-85](https://linear.app/elevenideas/issue/ATC-85/scaffold-the-app-server-package-serve-skeleton-and-ci): the production-shaped TypeScript App Server foundation, created alongside the existing Go `server/` (which is untouched).

## What's here

- **`app-server/`** — one Bun package, one `package.json`, committed `bun.lock`, Bun 1.3.14 pinned in `mise.toml`. Production deps are exactly Hono 4.12.32 and Commander 15.0.0, pinned.
- **CLI**: `atc serve` (Commander) runs the server in the foreground on loopback TCP. `--port` is strictly validated (digits only, ≤ 65535); the temporary dev default is 7332 so the Go server (7331) can run beside it.
- **Boundaries**: `src/main.ts` (entrypoint) → `src/cli/program.ts` (command registration/parsing) → `src/server/serve.ts` (foreground lifecycle) → `src/server/lifecycle.ts` (listener start/stop, shutdown signals) → `src/server/app.ts` (app factory, no listener needed) → `src/server/routes.ts` (typed `/api/v1` handlers) → `src/buildInfo.ts` (injected build metadata).
- **Endpoints**: `GET /api/v1/health` → `{"status":"ok"}`; `GET /api/v1/version` → version, `apiVersion`, commit, builtAt (camelCase, typed, no env reads in handlers).
- **Shutdown**: SIGINT/SIGTERM drain in-flight requests within a bounded grace window (5 s default, injectable in tests), then force-close — a hung request can never wedge Ctrl+C. Startup failures (e.g. port in use) print one line, not a stack trace.
- **Tests** (22): in-process app tests without a listener; listener bind/stop with port-release proof; forced-shutdown grace; signal listener cleanup; table-driven port validation; and a black-box test that spawns the real entrypoint on an ephemeral port, hits both endpoints over TCP, and asserts exit 0 on SIGTERM.
- **Workflow/CI**: mise tasks (`install`, `dev`, `fmt`, `fmt:check`, `typecheck`, `test`, `check`) fanned into root `check`/`test`; `app-server-ci.yml` runs the same `mise run check` (frozen-lockfile install, Prettier check, strict tsc, tests) mirroring the Server CI pattern.
- **Docs**: `app-server/README.md` (one command per workflow), root README table/task updates and stale `docs/` reference cleanup, App Server conventions recorded in `AGENTS.md`.

## Verification

- `mise run app-server:check` green locally (format, strict typecheck, 22 tests); suite ran repeatedly with zero flakes.
- Fresh-copy `bun install --frozen-lockfile` verified clean with the lockfile unmodified.
- Go server tests still pass; no changes under `server/`, `macos/`, or `packages/`.
- Two adversarial reviews (correctness, simplicity) were run; all accepted findings are incorporated (bounded shutdown grace, strict port parsing, friendly startup errors, e2e hardening, removal of unused seams/config).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,6 +16,32 @@ Long term maintainability is a core priority. If you add new functionality, firs
 
 The server stands on its own and is meant as a flexible resource that other apps can connect to and be built on top of. The Web UI is more admin interface than API client. It's meant as a place to manage all aspects of the server and document the CLI and API. The CLI and API should mirror each other in terms of functionality unless there is a good reason they should not.
 
+## App Server (app-server/)
+
+The TypeScript/Bun successor to the Go server. Conventions established there:
+
+- One Bun package: one `package.json`, one committed `bun.lock`. Install with
+  `bun install --frozen-lockfile` (the `install` mise task).
+- The Bun version is pinned exactly in `app-server/mise.toml`; no floating
+  channels.
+- Dependencies are pinned exactly and stay minimal. Hono and Commander are the
+  only runtime dependencies; adding another requires concrete justification.
+- Prettier is the single formatting solution; strict `tsc --noEmit` is the
+  type gate. No overlapping lint/format systems.
+- Boundaries: `src/main.ts` (entrypoint) → `src/cli/` (Commander registration
+  and parsing) → `src/server/` (serve lifecycle, listener start/stop, app
+  factory, routes) → `src/buildInfo.ts` (injected build metadata). The HTTP
+  app is constructible without a listener; build metadata, listener address,
+  and process lifecycle are injectable in tests.
+- Public HTTP routes are versioned under `/api/v1`; responses are typed JSON
+  with camelCase fields.
+- Tests live in `app-server/tests/` and run with `bun test`; they include a
+  black-box test that spawns the real entrypoint on an ephemeral loopback
+  port and verifies clean SIGTERM shutdown.
+- Workflows are mise tasks (`install`, `dev`, `fmt`, `fmt:check`, `typecheck`,
+  `test`, `check`); root `check`/`test` fan out to the package, and CI runs
+  the same `mise run check`.
+
 ## Source Control
 
 Jujutsu (jj) Protocol: You are in a jj repository; strictly do not use git
--- a/app-server/src/server/lifecycle.ts
+++ b/app-server/src/server/lifecycle.ts
@@ -1,0 +1,71 @@
+import type { Hono } from "hono";
+
+/** How long stop() waits for in-flight requests before force-closing. */
+const DEFAULT_SHUTDOWN_GRACE_MS = 5000;
+
+export interface ListenOptions {
+  hostname: string;
+  /** TCP port; 0 asks the OS for an ephemeral port. */
+  port: number;
+  /** Drain window for stop(); tests shorten it to keep hung-request coverage fast. */
+  shutdownGraceMs?: number;
+}
+
+export interface RunningServer {
+  /** The bound address, with the real port when an ephemeral one was requested. */
+  url: URL;
+  stop(): Promise<void>;
+}
+
+/** Bind the application to a TCP listener. */
+export function startServer(
+  app: Hono,
+  {
+    hostname,
+    port,
+    shutdownGraceMs = DEFAULT_SHUTDOWN_GRACE_MS,
+  }: ListenOptions,
+): RunningServer {
+  // Bun offers no bounded graceful stop: stop() waits on in-flight requests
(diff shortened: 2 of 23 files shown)


app-server/src/server/lifecycle.ts
```diff
@@ -0,0 +1,61 @@
...
+  }: ListenOptions,
+): RunningServer {
+  const server = Bun.serve({ hostname, port, fetch: app.fetch });
+  return {
+    url: server.url,
+    stop: async () => {
+      // Drain in-flight requests within the grace window, then close. A
+      // graceful stop() would hang forever on a request that never finishes,
+      // and Bun ignores stop(true) once a graceful stop is already pending.
+      const deadline = Date.now() + shutdownGraceMs;
```
jeremytondo, line 46: **[P1] Stop accepting new work before waiting for requests to drain.** This loop runs before any call to `server.stop`, so the listener remains live for the entire grace window. I reproduced this with a pending `/hang` request: a `/probe` request sent after `stop()` began was still accepted with 200. That means SIGTERM does not actually initiate listener shutdown, and new traffic can continue arriving until the eventual force-close. Since hanging handlers/SSE/WebSockets are explicit non-goals for ATC-85, the simplest correct skeleton is the documented `await server.stop()`; otherwise introduce an explicit draining/rejection mechanism and test that requests begun after shutdown are not accepted.

jeremytondo (commented):
The overall scaffold is small, well-factored, and matches most of ATC-85; the package check and hosted CI are green. I found one shutdown correctness issue that misses an explicit acceptance criterion, plus two smaller scope/test-quality issues.

app-server/tests/lifecycle.test.ts
```diff
@@ -0,0 +1,62 @@
...
+  hanging.get("/hang", () => new Promise<Response>(() => {}));
+
+  const server = startServer(hanging, {
+    hostname: "127.0.0.1",
+    port: 0,
+    shutdownGraceMs: 50,
+  });
+  const pending = fetch(new URL("/hang", server.url)).catch(() => "aborted");
+  // Let the request reach the server before initiating shutdown.
+  await Bun.sleep(20);
```
jeremytondo: **[P2] Synchronize with the handler instead of sleeping.** Twenty milliseconds does not prove `/hang` reached the server. If the fetch is still connecting, `stop()` can observe zero pending requests, close the listener, and make the fetch reject—so this test passes without exercising the `pendingRequests > 0` branch. Have the handler resolve an `entered` promise and await that before calling `stop()`.

AGENTS.md
```diff
@@ -50,6 +76,13 @@ When starting sub agents or running workflows, be smart about which agents to ch
 
 When doing work in Linear, always work within the atc team: https://linear.app/elevenideas/team/ATC.
 
+How work is organized:
+
+- An implementation issue is one PR. Its description carries everything needed to do the work: outcome, settled direction, scope, non-goals, and acceptance criteria.
```
jeremytondo: **[P2] Keep the ATC-85 agent guidance scoped to conventions established by this implementation.** These Linear workflow policies are not App Server structural conventions created or verified here, while the ticket explicitly says to update `AGENTS.md` only with structural conventions established and accepted in this PR. Please move this policy change to its own reviewable change or omit it from ATC-85.

app-server/src/server/lifecycle.ts
jeremytondo: Fixed. `stop()` now flips a `stopping` flag before draining: requests that arrive after shutdown begins get an immediate 503 instead of being served, in-flight requests drain within the grace window, then one `stop(true)` force-closes. The graceful-`server.stop()` route wasn't used because Bun ignores `stop(true)` while a graceful stop is pending (verified empirically on Bun 1.3.14 — a hung request only cleared via the 10s default idleTimeout), which would make shutdown unbounded. Added a test asserting a request begun during the drain window gets 503 while the original in-flight request is force-closed.

app-server/tests/lifecycle.test.ts
jeremytondo: Fixed. The hanging handler now resolves a `requestEntered` promise, and the test awaits it before calling `stop()`, so the `pendingRequests > 0` branch is provably exercised. The sleep is gone.

AGENTS.md
jeremytondo: Fixed. The Linear workflow / README housekeeping commit was rebased out of this branch and moved to its own reviewable change: #122. This PR now contains only the ATC-85 commit, and its AGENTS.md diff is limited to the App Server conventions section.

PR #122, opened 2026-07-31
Refactor prep: drop stale docs/ references from README

jeremytondo:
Split out of #121 per review: this README housekeeping change removes stale `docs/` references. It is not an App Server convention established by ATC-85, so it gets its own reviewable change.

https://claude.ai/code/session_01Hc5f2vXdqm3Un6J4GfJ4qp


AGENTS.md
```diff
@@ -50,6 +50,13 @@ When starting sub agents or running workflows, be smart about which agents to ch
 
 When doing work in Linear, always work within the atc team: https://linear.app/elevenideas/team/ATC.
 
+How work is organized:
```
jeremytondo: I don't think we need to add this how work is organized section. Please remove

jeremytondo: Removed the entire “How work is organized” section. The PR now only contains the stale `docs/` reference cleanup in `README.md`.

PR #123, opened 2026-07-31
ATC-117: Scaffold the Effect App Server package, serve skeleton, and CI

jeremytondo:
Implements [ATC-117](https://linear.app/elevenideas/issue/ATC-117/scaffold-the-effect-app-server-package-serve-skeleton-and-ci): the production-shaped TypeScript App Server foundation, built on **Effect** as the application spine. Replaces #121 (Hono + Commander scaffold), which will be abandoned.

## What's here

- **`app-server/`** — one Bun + strict TypeScript package. Production deps are exactly two, exact-pinned: `effect@4.0.0-beta.102` and `@effect/platform-bun@4.0.0-beta.102` (the same line T3Code and OpenCode build on). Committed `bun.lock`; Bun 1.3.14 pinned in `app-server/mise.toml`.
- **Contract** (`src/api.ts`) — the `/api/v1` group (`health`, `version`) as a schema-first `HttpApi` module with Schema-typed camelCase responses. The server implementation, OpenAPI document, and future typed clients all derive from this one definition (OpenAPI serving itself is follow-up work).
- **CLI + lifecycle** — `atc serve` via Effect CLI: foreground loopback listener, `--port` validated 1–65535 (temporary dev default 7332 so the Go server on 7331 runs alongside). Graceful SIGINT/SIGTERM shutdown flows through `BunRuntime.runMain` interruption and scoped finalizers — no hand-rolled signal listeners. Startup failures with a syscall code (e.g. `EADDRINUSE`) surface as one friendly stderr line and exit 1; anything else re-dies so real bugs keep the loud report.
- **Build metadata** (`src/buildInfo.ts`) — `BuildInfo` as an injected `Context.Service` (version from package.json; commit/builtAt are dev placeholders until the standalone-executable build injects real values).
- **Tests** (`@effect/vitest`, run on the Bun runtime via `bun --bun vitest`) — in-process contract tests through `HttpApiTest` with a substituted BuildInfo layer (no listener); real-listener bind/serve/port-release; `--port` validation down to the wrapped `InvalidValue`; and black-box tests that spawn the real entrypoint, hit both endpoints over TCP, hold a half-sent request open across the signal, and assert clean exit 130 on both SIGTERM and SIGINT plus the one-line startup failure.
- **Workflow/CI** — app-server mise tasks (`install`/`dev`/`fmt`/`fmt:check`/`typecheck`/`test`/`check`) fanned into root `check`/`test`; App Server CI workflow runs the same `mise run check` (frozen-lockfile install, Prettier check, strict tsc, vitest).
- **Conventions** — Effect conventions section added to `AGENTS.md` (Context.Service + Layer shape, tagged errors only, one `runMain`, `it.effect`, Schema for validation, exact pins, reference checkouts).

## Key decisions

- **Effect v4 API reality check**: on the 4.0.0-beta line everything lives in the single `effect` package (`effect/unstable/httpapi`, `effect/unstable/cli` where v3's `Options` is now `Flag`); serving is `HttpRouter.serve(HttpApiBuilder.layer(Api))` + `BunHttpServer.layer` (v3's `HttpApiBuilder.serve` no longer exists). Verified against the pinned checkout (`mise run refs`), T3Code, and OpenCode rather than training data.
- **Tests run on Bun**: `bun --bun vitest` forces the Bun runtime (vitest's shebang would pick Node), so tests exercise the same runtime the server ships on — verified `BunHttpServer` works in-process under it.
- **Friendly startup errors**: bind failures are defects (Bun.serve throws), caught only when they carry a syscall `code`, printed as one line, and marked `[Runtime.errorReported] = false` so `runMain` doesn't re-dump the cause. Non-system defects re-die.
- **Adversarially reviewed**: two independent review agents (correctness, simplicity) ran against the scaffold; accepted findings are in the second commit (`Layer.launch` over hand-rolled scope/build/never, narrowed defect catch, real in-flight/SIGINT black-box coverage, `it.live` for socket tests, dropped premature build-define machinery, deleted redundant `.gitignore`).

The two small prep commits (TypeScript `.gitignore` defaults, `mise run refs` for read-only reference checkouts) ride along at the base of the branch.

## Verification

- `mise run check` (app-server): Prettier check, strict tsc, 12 tests — green; suite stress-run 5× with no flakes.
- `mise run server:check`: Go server unaffected — green.
- Manual: `atc serve` on 7399, both endpoints, Ctrl+C/SIGTERM clean shutdown with port release; port-in-use prints exactly one line.

https://claude.ai/code/session_01Hc5f2vXdqm3Un6J4GfJ4qp

### DIFF
--- a/app-server/README.md
+++ b/app-server/README.md
@@ -1,9 +1,7 @@
 # ATC App Server
 
 The TypeScript implementation of the ATC server and `atc` CLI, built on
-[Effect](https://effect.website) and [Bun](https://bun.sh). It lives alongside
-the existing Go server in [`../server/`](../server/) until cutover; the Go
-server remains the installed product today.
+[Effect](https://effect.website) and [Bun](https://bun.sh).
 
 Bun is pinned in [`mise.toml`](mise.toml) and installed automatically by
 [mise](https://mise.jdx.dev). All workflows are mise tasks:
@@ -18,17 +16,15 @@ mise run typecheck  # strict TypeScript type check
 mise run check      # everything CI runs: fmt:check + typecheck + test
 ```
 
-The dev port 7332 is a temporary development default (the Go server uses
-7331); production listener behavior is settled in later work.
+The server listens on port 7332 by default; pass `--port` to override.
 
 ## Structure
 
 One Bun package, one `package.json`, one committed `bun.lock`. Runtime
 dependencies are pinned exactly and kept minimal: `effect` (the application
 spine — HTTP, CLI, Schema, concurrency) and `@effect/platform-bun` (the Bun
-adapter). Effect is on the `4.0.0-beta.x` line, upgraded deliberately, never
-floated. `mise run refs` (repo root) checks out the matching Effect source for
-API reference.
+adapter), upgraded deliberately, never floated. `mise run refs` (repo root)
+checks out the matching Effect source for API reference.
 
 | Path               | Responsibility                                                    |
 | ------------------ | ----------------------------------------------------------------- |
--- a/app-server/README.md
+++ b/app-server/README.md
@@ -1,0 +1,50 @@
+# ATC App Server
+
+The TypeScript implementation of the ATC server and `atc` CLI, built on
+[Effect](https://effect.website) and [Bun](https://bun.sh). It lives alongside
+the existing Go server in [`../server/`](../server/) until cutover; the Go
+server remains the installed product today.
+
+Bun is pinned in [`mise.toml`](mise.toml) and installed automatically by
+[mise](https://mise.jdx.dev). All workflows are mise tasks:
+
+```sh
+mise run install    # install dependencies from the committed bun.lock
+mise run dev        # run `atc serve` in the foreground (http://127.0.0.1:7332)
+mise run test       # run the vitest suite on the Bun runtime
+mise run fmt        # format with Prettier
+mise run fmt:check  # fail if formatting is needed
+mise run typecheck  # strict TypeScript type check
+mise run check      # everything CI runs: fmt:check + typecheck + test
+```
+
+The dev port 7332 is a temporary development default (the Go server uses
+7331); production listener behavior is settled in later work.
+
+## Structure
+
+One Bun package, one `package.json`, one committed `bun.lock`. Runtime
+dependencies are pinned exactly and kept minimal: `effect` (the application
+spine — HTTP, CLI, Schema, concurrency) and `@effect/platform-bun` (the Bun
+adapter). Effect is on the `4.0.0-beta.x` line, upgraded deliberately, never
+floated. `mise run refs` (repo root) checks out the matching Effect source for
+API reference.
+
+| Path               | Responsibility                                                    |
+| ------------------ | ----------------------------------------------------------------- |
+| `src/main.ts`      | Entrypoint for the `atc` executable; the only `runMain`           |
+| `src/cli.ts`       | CLI commands and flags (Effect CLI); friendly startup failures    |
+| `src/api.ts`       | The `/api/v1` HttpApi contract: endpoints and response schemas    |
+| `src/handlers.ts`  | Contract implementation (handler Layer, no listener)              |
+| `src/server.ts`    | Server assembly: routes + loopback Bun listener Layer             |
+| `src/buildInfo.ts` | Build metadata service (version, commit, builtAt)                 |
+| `test/`            | `@effect/vitest` tests, including a black-box serve/shutdown test |
+
+The contract module is structured so the server implementation, an OpenAPI
+document, and generated typed clients all derive from the same `HttpApi`
+definition; OpenAPI serving and clients are follow-up work.
+
+Public HTTP routes are versioned under `/api/v1`:
+
+- `GET /api/v1/health` → `{ "status": "ok" }`
+- `GET /api/v1/version` → application version, `apiVersion`, and build metadata
--- a/app-server/mise.toml
+++ b/app-server/mise.toml
@@ -1,0 +1,44 @@
+[tools]
+bun = "1.3.14"
+
+# --- Primary entrypoints ---
+
+[tasks.dev]
+description = "Run the App Server in the foreground (temporary dev default: http://127.0.0.1:7332)"
+depends = ["install"]
+run = "bun run src/main.ts serve"
+
+[tasks.check]
+description = "All App Server gates: format check, strict type check, tests"
+depends = ["fmt:check", "typecheck", "test"]
+
+# --- Building blocks ---
+
+# fmt:check/typecheck/test share install as a dependency instead of each
+# installing themselves: mise runs tasks in parallel, and concurrent installs
+# into the same node_modules race. A shared dependency runs exactly once.
+[tasks.install]
+description = "Install dependencies from the committed lockfile"
+run = "bun install --frozen-lockfile"
+
+[tasks.fmt]
+description = "Format TypeScript, JSON, and Markdown with Prettier"
+depends = ["install"]
+run = "bun run prettier --write ."
+
+[tasks."fmt:check"]
+description = "Fail if any file is not Prettier-clean"
+depends = ["install"]
+run = "bun run prettier --check ."
+
+[tasks.typecheck]
+description = "Strict TypeScript type check (no emit)"
+depends = ["install"]
+run = "bun run tsc --noEmit"
+
+# --bun forces the Bun runtime (vitest's shebang would otherwise pick node),
+# so tests exercise the same runtime the server ships on.
+[tasks.test]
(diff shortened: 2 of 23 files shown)


app-server/src/api.ts
```diff
@@ -0,0 +1,32 @@
+import { Schema } from "effect"
+import { HttpApi, HttpApiEndpoint, HttpApiGroup, OpenApi } from "effect/unstable/httpapi"
+import { buildInfo } from "./buildInfo.ts"
```
jeremytondo: **[P2] Keep the reusable contract independent of the production BuildInfo module.** Importing `buildInfo.ts` makes every future typed-client/OpenAPI consumer of `api.ts` transitively load the server's `Context.Service`, Layer, and package metadata. That conflicts with the client-safe contract boundary described immediately below (and with the pinned Effect guide's explicit API/implementation separation), so the contract will need restructuring as soon as it is shared. Please keep the OpenAPI version contract-owned (for example `"v1"`), omit it until the build source exists, or move the immutable version into a client-safe module; only handlers should depend on the injected `BuildInfo` service.

app-server/src/main.ts
```diff
@@ -0,0 +1,13 @@
+import { BunRuntime, BunServices } from "@effect/platform-bun"
+import { Effect, Layer } from "effect"
+import { Command } from "effect/unstable/cli"
+import * as BuildInfo from "./buildInfo.ts"
+import { atc } from "./cli.ts"
+
+// The only runMain in the application. Everything else is Layers and Effects.
+// Command.run needs a plain string before any Layer exists, so it reads the
+// buildInfo constant directly; everything downstream uses the injected service.
+Command.run(atc, { version: BuildInfo.buildInfo.version }).pipe(
```
jeremytondo: **[P2] Resolve the CLI version through the injected BuildInfo service.** `BuildInfo.layer` cannot affect the value already read here, so substituting a BuildInfo Layer changes `/api/v1/version` but not `atc --version`; the comment that downstream code uses the service is therefore not true. `Command.run` can be invoked inside `Effect.gen`: yield `BuildInfo.BuildInfo`, then yield `Command.run(atc, { version: build.version })`, and provide the Layers to that Effect before the single `runMain`. That preserves the simple one-runtime shape while making the DI boundary consistent and testable.

app-server/src/cli.ts
```diff
@@ -0,0 +1,45 @@
+import { Console, Effect, Layer, Runtime, Schema } from "effect"
+import { Command, Flag } from "effect/unstable/cli"
+import * as Server from "./server.ts"
+
+// Temporary dev default so the Go server on 7331 can run alongside.
+export const DEFAULT_PORT = 7332
+
+export const port = Flag.integer("port").pipe(
+  Flag.withDefault(DEFAULT_PORT),
+  Flag.filter(
```
jeremytondo: **[P2] Put the port constraint in Effect Schema.** This scaffold establishes “all validation and domain types are Effect Schema,” but its first bounded domain value is implemented with a standalone `Flag.filter` predicate. Effect v4 provides `Flag.withSchema`, and the pinned references use `Schema.Int.check(Schema.isBetween({ minimum: 1, maximum: 65535 }))` for ports. Define that schema once and apply it to `Flag.integer`; this keeps the stated convention honest and gives later CLI/config/API inputs one reusable port contract.

jeremytondo (commented):
Reviewed the full diff against ATC-117, the replacement context from #121, and the pinned Effect 4.0.0-beta.102 source plus the T3Code/OpenCode references. The implementation is compact, the lifecycle wiring follows Effect's scoped Layer model, and both hosted CI and the local app-server check are green. I found three maintainability issues in the boundaries/conventions this scaffold is establishing; details are inline.

app-server/src/api.ts
jeremytondo: Fixed in 3b7cc9a: the contract no longer imports `buildInfo.ts`; the OpenAPI annotation carries the contract version (`"v1"`) with a comment noting it is deliberately not the build version, so `api.ts` stays client-safe.

app-server/src/main.ts
jeremytondo: Fixed in 3b7cc9a: the entrypoint is now a single `Effect.gen` that yields `BuildInfo.BuildInfo` and passes `build.version` to `Command.run`, with layers provided to that program before the one `runMain`. Substituting a BuildInfo layer now changes `atc --version` too.

app-server/src/cli.ts
jeremytondo: Fixed in 3b7cc9a: added `export const Port = Schema.Int.check(Schema.isBetween({ minimum: 1, maximum: 65535 }))` in cli.ts and applied it with `Flag.withSchema(Port)`, replacing the `Flag.filter` predicate. CLI error output still renders one clear line and the validation tests still pass unchanged.

app-server/README.md
```diff
@@ -0,0 +1,50 @@
+# ATC App Server
+
+The TypeScript implementation of the ATC server and `atc` CLI, built on
+[Effect](https://effect.website) and [Bun](https://bun.sh). It lives alongside
+the existing Go server in [`../server/`](../server/) until cutover; the Go
+server remains the installed product today.
```
jeremytondo: I think we can get rid of comments here that refer to the old Go codebase. Eventually that codebase will be gone and these comments will be odd and stale in the new code. Change this here and everywhere else we do this.

app-server/README.md
```diff
@@ -0,0 +1,50 @@
...
+mise run dev        # run `atc serve` in the foreground (http://127.0.0.1:7332)
+mise run test       # run the vitest suite on the Bun runtime
+mise run fmt        # format with Prettier
+mise run fmt:check  # fail if formatting is needed
+mise run typecheck  # strict TypeScript type check
+mise run check      # everything CI runs: fmt:check + typecheck + test
+```
+
+The dev port 7332 is a temporary development default (the Go server uses
+7331); production listener behavior is settled in later work.
```
jeremytondo: Not sure why this needs to be temporary, let's just make permananent. I can't see a good reason to change it after we switch over to the new app server.

app-server/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Structure
+
+One Bun package, one `package.json`, one committed `bun.lock`. Runtime
+dependencies are pinned exactly and kept minimal: `effect` (the application
+spine — HTTP, CLI, Schema, concurrency) and `@effect/platform-bun` (the Bun
+adapter). Effect is on the `4.0.0-beta.x` line, upgraded deliberately, never
```
jeremytondo: We should be careful about referring to specific version numbers here as that could get stale fast.

app-server/mise.toml
```diff
@@ -0,0 +1,44 @@
+[tools]
+bun = "1.3.14"
+
+# --- Primary entrypoints ---
+
+[tasks.dev]
+description = "Run the App Server in the foreground (temporary dev default: http://127.0.0.1:7332)"
```
jeremytondo: Why is this a temporary default? See my other comments about this.

app-server/src/cli.ts
```diff
@@ -0,0 +1,45 @@
+import { Console, Effect, Layer, Runtime, Schema } from "effect"
+import { Command, Flag } from "effect/unstable/cli"
+import * as Server from "./server.ts"
+
+// Temporary dev default so the Go server on 7331 can run alongside.
```
jeremytondo: See my other comments about why this likely doesn't need to be called out as temporary.

app-server/README.md
jeremytondo: Fixed in d48ab38: dropped the "lives alongside the existing Go server" sentence from the README intro, the Go-server aside in the port paragraph, and the Go-server comment in cli.ts. A sweep of every file the PR touches found no other Go-codebase references.

app-server/README.md
jeremytondo: Fixed in d48ab38: 7332 is now stated as the permanent default — "The server listens on port 7332 by default; pass `--port` to override." No temporary framing anywhere.

app-server/README.md
jeremytondo: Fixed in d48ab38: removed the `4.0.0-beta.x` version string from the README; it now just says the pins are exact and upgraded deliberately. The actual versions live only in package.json/bun.lock.

app-server/mise.toml
jeremytondo: Fixed in d48ab38: the dev task description is now just "Run the App Server in the foreground (http://127.0.0.1:7332)".

app-server/src/cli.ts
jeremytondo: Fixed in d48ab38: deleted the comment entirely — `DEFAULT_PORT = 7332` stands on its own now that it is the permanent default.

PR #124, opened 2026-08-01
ATC-88: Prove standalone compilation and provider spawning

jeremytondo:
Implements [ATC-88](https://linear.app/elevenideas/issue/ATC-88/prove-standalone-compilation-and-provider-spawning): proves the production-shaped `atc` executable compiles with Bun and can launch both provider technologies, before the App Server grows around assumptions that don't hold in a compiled executable.

## What's here

**Standalone executable**
- `mise run build` / `mise run build:all` (fanned into root tasks) compile `src/main.ts` into deterministic `dist/atc-<os>-<arch>` artifacts for darwin/linux × arm64/x64.
- Build metadata is injected at compile time via `--define` (commit + `-dirty` marker, build timestamp); source runs report `dev`. `.env`/bunfig autoloading is disabled so compiled behavior never depends on the working directory.

**Generic subprocess seam** (`src/subprocess.ts`)
- `Subprocess` service on Effect's `ChildProcess`/`ChildProcessSpawner`: scoped acquisition (scope close terminates the child, SIGTERM → SIGKILL escalation), explicit env, line-oriented stdin/stdout, bounded stderr diagnostics, stdin writes that fail loudly once the child is gone.

**Provider smoke checks** (`src/smoke.ts`, hidden `atc smoke codex|claude`)
- Both providers follow one bring-your-own-CLI rule: atc ships no provider binaries. Resolution is explicit — env override (`ATC_CODEX_EXECUTABLE` / `ATC_CLAUDE_CODE_EXECUTABLE`) → the user's install on PATH.
- Codex: one JSON-RPC `initialize`/`initialized` round trip over stdio, clean shutdown, explicit no-orphan verification.
- Claude: Agent SDK round trip with `pathToClaudeCodeExecutable` always set explicitly (the SDK's node_modules-based default resolution is dead inside a compiled binary); runs from a temp cwd outside any repo; orphan check runs on success *and* failure.
- The command is `Command.withHidden`, marked unstable, and will be replaced by the production provider integration.

**Tests & CI**
- Fixture-based seam tests (launch, capture, timeout, cancellation, exit, cleanup, SIGKILL escalation, stdin-after-exit) and a fake codex app-server round-trip test run in the default fast suite — no provider accounts needed.
- Opt-in `mise run test:compiled` (compiled artifact black-box: health/version with real metadata from outside the repo, clean SIGTERM) and `mise run test:smoke` (live providers, prerequisite checks).
- CI: cross-compile job (build-only, labeled as such) + native compiled validation on macOS arm64 and Linux x64, which also run the OS-sensitive fast suite natively.
- AGENTS.md records the compiled-asset/executable-discovery/subprocess conventions and the Bun/Effect upgrade re-validation requirement.

## Key decisions

- **Claude Code is the user's install, not shipped by atc.** The SDK's packaged platform binary is ~245 MB per target and resolves via node_modules — a dead path inside a compiled executable. An earlier revision of this PR staged that binary next to the artifact; after comparing with T3Code's approach we settled on relying on the user's installed `claude` instead: the user must have an authenticated Claude Code install anyway (the binary is useless without its credentials), it makes both providers symmetric, and it keeps artifacts at ~65–100 MB with no per-target binary acquisition at release time. This deliberately supersedes the issue's original "packaged platform-specific executable" direction. Version-compatibility advisories (user's CLI vs pinned SDK) are follow-up work for the production adapter.
- **Codex round trip = `initialize`.** A real structured request/response over the app-server protocol that needs no thread state and leaves nothing behind in `~/.codex`.
- **Live smokes verified locally**: both `atc smoke codex` and `atc smoke claude` pass from the compiled binary running outside the repository (codex-cli 0.146.0, Agent SDK 0.3.220 driving the user's installed Claude Code 2.1.220).

Two adversarial reviews (correctness, simplicity) ran before this PR; all major findings were applied (stdin write contract, always-on orphan check, dirty marker, JSON tolerance on provider stdout, native fast suite in CI).

### DIFF
--- a/app-server/src/smoke.ts
+++ b/app-server/src/smoke.ts
@@ -282,8 +282,7 @@ export const claudeSmoke = (claudeExecutable: string) =>
       yield* Effect.addFinalizer(() => Effect.sync(() => abortController.abort()))
       const summary = yield* Effect.tryPromise({
         try: () => claudeRoundTrip(claudeExecutable, cwd, abortController),
-        catch: (error) =>
-          new SmokeError({ provider: "claude", message: describeError(error) }),
+        catch: (error) => new SmokeError({ provider: "claude", message: describeError(error) }),
       }).pipe(
         Effect.timeoutOrElse({
           duration: "180 seconds",
--- a/app-server/src/smoke.ts
+++ b/app-server/src/smoke.ts
@@ -1,11 +1,9 @@
 import { query, type SDKMessage } from "@anthropic-ai/claude-agent-sdk"
 import { Console, Deferred, Duration, Effect, Runtime, Schema, Stream } from "effect"
 import { Command } from "effect/unstable/cli"
-import { existsSync } from "node:fs"
 import { mkdtemp, rm } from "node:fs/promises"
-import { createRequire } from "node:module"
 import { tmpdir } from "node:os"
-import { dirname, join } from "node:path"
+import { join } from "node:path"
 import * as BuildInfo from "./buildInfo.ts"
 import * as Subprocess from "./subprocess.ts"
 
@@ -42,31 +40,22 @@ export const resolveCodexExecutable = Effect.suspend(() => {
 })
 
 /**
- * Claude Code: explicit env override, then the packaged platform binary staged
- * next to this executable (compiled distribution; target-scoped name so a
- * wrong-platform binary can never pair), then the platform package in
- * node_modules (running from source). Never the working directory.
+ * Claude Code: explicit env override, then the user's installed `claude` on
+ * PATH — the same bring-your-own-CLI rule as codex. atc does not ship a
+ * Claude Code binary; the user's install also carries the credentials the
+ * SDK needs.
  */
 export const resolveClaudeCodeExecutable = Effect.suspend(() => {
   const override = process.env["ATC_CLAUDE_CODE_EXECUTABLE"]
   if (override !== undefined && override !== "") return Effect.succeed(override)
-  const platformTarget = `${process.platform}-${process.arch}`
-  const adjacent = join(dirname(process.execPath), `claude-${platformTarget}`)
-  if (existsSync(adjacent)) return Effect.succeed(adjacent)
-  const platformPackage = `@anthropic-ai/claude-agent-sdk-${platformTarget}`
-  try {
-    const resolved = createRequire(import.meta.url).resolve(`${platformPackage}/claude`)
-    if (existsSync(resolved)) return Effect.succeed(resolved)
-  } catch {
-    // No node_modules in a compiled executable; fall through to the error.
-  }
+  const found = Bun.which("claude")
+  if (found !== null) return Effect.succeed(found)
   return Effect.fail(
     new SmokeError({
       provider: "claude",
       message:
-        "Claude Code executable not found; tried $ATC_CLAUDE_CODE_EXECUTABLE, " +
-        `${adjacent}, and ${platformPackage} in node_modules. ` +
-        "Run `mise run build` to stage it, or set ATC_CLAUDE_CODE_EXECUTABLE.",
+        "claude not found on PATH; install and authenticate Claude Code " +
+        "or set ATC_CLAUDE_CODE_EXECUTABLE",
     }),
   )
 })
--- a/app-server/src/smoke.ts
+++ b/app-server/src/smoke.ts
@@ -1,5 +1,5 @@
 import { query, type SDKMessage } from "@anthropic-ai/claude-agent-sdk"
-import { Console, Duration, Effect, Option, Runtime, Schema, Stream } from "effect"
+import { Console, Deferred, Duration, Effect, Runtime, Schema, Stream } from "effect"
 import { Command } from "effect/unstable/cli"
 import { existsSync } from "node:fs"
 import { mkdtemp, rm } from "node:fs/promises"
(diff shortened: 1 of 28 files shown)


app-server/src/smoke.ts
```diff
@@ -0,0 +1,412 @@
...
+        )
+      }
+      yield* Effect.sleep("100 millis")
+    }
+  })
+
+// --- Codex: one JSON-RPC initialize round trip over stdio --------------------
+
+/** The one message shape this smoke check inspects; `error: null` counts as absent. */
+const InitializeResponse = Schema.Struct({
```
jeremytondo, line 138: [P2] Require an actual initialize result before accepting the response

Every field in this schema is optional, so a matching response such as `{ "id": 1 }` decodes successfully and falls through to `initialize round trip OK`. That lets a broken or protocol-incompatible app-server pass the smoke check without returning an initialize result. Please model this as a success/error union that requires `result` on the success branch (ideally the expected `userAgent`), and add a fixture case for a response with neither `result` nor `error`.

app-server/src/smoke.ts
```diff
@@ -0,0 +1,412 @@
...
+          )
+        }
+        yield* Console.log("initialize round trip OK")
+        yield* child.writeLine(JSON.stringify({ method: "initialized", params: {} }))
+        yield* child.endInput
+        const exitCode = yield* child.exitCode.pipe(
+          Effect.timeoutOrElse({ duration: "5 seconds", orElse: () => Effect.succeed(null) }),
+          Effect.orElseSucceed(() => null),
+        )
+        yield* exitCode === null
```
jeremytondo, line 278: [P2] Fail the smoke check when Codex exits non-zero

A child that answers initialize and then exits with code 1 currently logs `exited with code 1` and continues to the final `PASS`. Since this check claims a clean shutdown, any non-zero numeric exit should fail with the captured stderr; only the timeout/null branch should be treated as an intentional forced cleanup. A fixture mode that exits non-zero after responding would lock this down.

app-server/src/smoke.ts
```diff
@@ -0,0 +1,412 @@
...
+      const survivors = yield* Effect.scoped(
+        Effect.gen(function* () {
+          const pgrep = yield* subprocess.spawn({
+            executable: "pgrep",
+            args: ["-l", "-P", String(process.pid)],
+            env: {},
+            extendEnv: true,
+          })
+          const lines = yield* Stream.runCollect(pgrep.stdoutLines)
+          yield* pgrep.exitCode // 0 = matches, 1 = none; either is a clean run
```
jeremytondo: [P2] Reject `pgrep` exit codes other than 0 or 1

`Child.exitCode` returns non-zero codes as values, so this line silently accepts exit 2+ as well. If `pgrep` cannot inspect the process table or rejects its arguments and prints no stdout, `survivors` is empty and the code declares the orphan check successful. Please inspect the code explicitly: accept 0 (matches) and 1 (none), but turn anything else into `SmokeError` with `stderrTail` diagnostics.

app-server/src/smoke.ts
```diff
@@ -0,0 +1,412 @@
...
+                }),
+              ),
+          }),
+        ),
+      )
+      // Whatever happened, the SDK's child must be gone before we finish
+      // (and before the temp cwd is removed underneath it). Keep the original
+      // failure primary so an orphan report never masks the root cause.
+      abortController.abort()
+      const orphanCheck = yield* Effect.result(verifyNoChildOrphans("claude"))
```
jeremytondo, line 377: [P2] Wait through the SDK's termination escalation before declaring cleanup failed

In the pinned Agent SDK, abort closes stdin, waits 2 seconds before sending SIGTERM, then waits another 5 seconds before SIGKILL; those escalation timers are unref'ed. This orphan loop also expires at roughly 2 seconds, so a timed-out/failing query can reach this line just before the SDK sends its first signal, report an orphan, and let `atc` exit before the SDK's SIGKILL timer can run. Please make the cleanup budget exceed the SDK's full escalation window or take ownership of spawning/waiting through the SDK hook so the child is definitely reaped before the temp cwd and parent process go away.

app-server/src/subprocess.ts
```diff
@@ -0,0 +1,172 @@
...
+  yield* handle.exitCode.pipe(Effect.ignore, Effect.andThen(Queue.end(stdin)), Effect.forkScoped)
+
+  // Drain stderr concurrently into a bounded tail so a chatty child cannot
+  // fill the pipe (deadlock) or grow diagnostics without limit. Note the
+  // bound is per line after splitting: output with no newlines at all still
+  // accumulates in splitLines before truncation applies.
+  const stderrTail: Array<string> = []
+  yield* handle.stderr.pipe(
+    Stream.decodeText,
+    Stream.splitLines,
```
jeremytondo: [P2] Enforce the stderr bound before line splitting

`Stream.splitLines` must retain an incomplete line, so a child that continuously writes stderr without `\n` can grow this fiber without limit and OOM the server before `truncate` ever runs. That contradicts the bounded-diagnostics contract this service advertises (and the comment already notes the hole). Please cap bytes/chunks before or during incremental line decoding, and cover a large no-newline stderr fixture.

jeremytondo (commented):
Full review completed against `66ceb70b`. The overall structure and test coverage are strong, and the fast suite plus compiled black-box test pass locally. I found five correctness/reliability gaps that can make the smoke checks report success without proving their stated postconditions, or leave diagnostics/process cleanup unbounded.

app-server/src/smoke.ts
jeremytondo: Fixed in d2e9f843: `InitializeResponse` now models `result` explicitly, and a response with neither a non-null `result` object nor an `error` fails with "initialize response carried neither result nor error" plus stderr diagnostics. Added an "empty" fixture mode ({id} only) with a test locking it down. Kept the requirement at "result is an object" rather than asserting `userAgent` specifically, so the smoke check doesn't couple to a field the protocol may rename across codex versions.

app-server/src/smoke.ts
jeremytondo: Fixed in d2e9f843: a non-zero numeric exit after the round trip now fails the smoke with the captured stderr ("exited with code N after the round trip"); only the null timeout branch remains the deliberate forced-cleanup path. Added an "exit-nonzero" fixture mode + test.

app-server/src/smoke.ts
jeremytondo: Fixed in d2e9f843: pgrep exit codes are now inspected explicitly — 0 and 1 are clean runs, anything else fails with "orphan check could not run: pgrep exited with N" plus pgrep's stderr tail.

app-server/src/smoke.ts
jeremytondo: Fixed in d2e9f843. Verified the escalation in the pinned sdk.mjs (stdin close → 2s → SIGTERM → 5s → SIGKILL, all unref'ed): the settle window is now 100 × 100ms ≈ 10s, comfortably past the ~7s escalation, with a comment explaining why the window must outlast it (our exit kills the SDK's unref'ed timers). Owning the spawn via the SDK hook is the better long-term shape but belongs to the production adapter work.

app-server/src/subprocess.ts
jeremytondo: Fixed in d2e9f843: stderr line-splitting is now hand-rolled in the drain fiber — the unterminated fragment is hard-capped at MAX_CAPTURED_LINE_LENGTH+1, so a child writing without newlines can no longer grow memory; the eventual line is marked truncated. Covered by a 2MB no-newline fixture test. (stdoutLines keeps Stream.splitLines — it's the caller's protocol stream, documented single-consumer.)

PR #127, opened 2026-08-02
ATC-121: Core persistence, configuration, diagnostics, and local trust

jeremytondo:
Implements [ATC-121](https://linear.app/elevenideas/issue/ATC-121/core-persistence-configuration-diagnostics-and-authentication): the durable control-plane substrate — settled configuration and data locations, transactional SQLite persistence with checked-in migrations, a structured logging baseline, a hardened loopback listener, and the first domain resource (Projects) end to end through the contract, clients, and CLI.

## What's here (one commit per area)

- **Configuration** (`src/config.ts`): one precedence rule — flags > environment (flat `ATC_<KEY>`) > `config.toml` (Bun-native TOML, camelCase keys, unknown keys rejected) > defaults — via composed `ConfigProvider`s. One XDG path rule on every platform (`~/.config/atc/config.toml`, `~/.local/share/atc/atc.db`, `~/.local/state/atc/atc.log`), honoring `XDG_*` overrides plus `ATC_CONFIG`/`ATC_DATA_DIR`. Invalid config fails fast with one stderr line naming the offending source; relative data/config paths are rejected so compiled behavior never depends on the cwd.
- **Persistence** (`src/persistence.ts`, `src/migrations.ts`): `@effect/sql-sqlite-bun@4.0.0-beta.102` (exact-pinned, same line as `effect`). In-code append-only `Migrator.fromRecord` record (`"0001_init"`), applied transactionally at startup with library bookkeeping and concurrent-run locking; a failed migration rolls back DDL+bookkeeping together and halts boot naming the migration. Documented pragmas: WAL (driver default), `busy_timeout=5000`, `foreign_keys=ON`. A test pins loader count to record size (`fromRecord` silently drops malformed keys).
- **Logging** (`src/logging.ts`): JSON-lines file in the state dir + pretty stderr in dev (`commit === "dev"`), level from `ATC_LOG_LEVEL`/config (case-insensitive), request spans via `HttpMiddleware.tracer` with trace/span ids annotated onto handler logs.
- **Listener hardening** (`src/localTrust.ts`): loopback-only bind plus `Host`/`Origin` validation on every request — spoofed hosts (DNS rebinding) and cross-origin browser requests (CSRF) get an empty 403. Covered by unit predicates, real-listener tests (raw-socket Host spoofing), and the compiled black-box suite.
- **Projects + fs.check** (`src/api.ts`, `src/handlers.ts`, `src/projectRepository.ts`, `src/directories.ts`): full CRUD per the Domain Model — UUIDv7 ids, required mutable `defaultWorkingDirectory` stored symlink-resolved canonical, validation requiring an existing readable+traversable directory (deliberately not writable), tagged errors with `httpApiStatus` (404/422) and human messages, simple record delete. `GET /api/v1/fs/check?path=` returns the tagged health states with a 2s bounded timeout (`unknown`/timeout for explicit checks; fail-closed retryable error inside operations); never persisted. Row types stay behind the repository.
- **CLI** (`src/cli.ts`): `--url` is gone — API-backed commands derive `http://127.0.0.1:<port>` from the same settled config the server reads (single resolution seam kept for the future remote/auth work). New `atc project list|get|create|update|delete` (delete requires `--yes`) and `atc fs check`; relative directories resolve client-side. JSON stdout / one-line stderr diagnostics / exit-code contract preserved and black-box tested.
- **Contract artifacts**: `openapi.json` regenerated; Swift client (`ATCAppServerAPI`) builds and its tests pass via `mise run contract:check`.
- **Docs**: README + AGENTS.md document the settled locations, precedence rule, persistence conventions, and local-trust architecture.

## Notable decisions

- `createProject` returns **200 with the body**, not 201: in this Effect version a per-endpoint status annotation on a shared named schema forks it into a duplicate OpenAPI component (and a duplicate generated Swift type). The archived contract doc's success-code table predates the Effect stack decision.
- Known contract-polish gap (deliberate, follow-up material): generated request schemas say `additionalProperties: false` but the runtime ignores excess JSON keys, and schema-validation 400s (empty body) are undocumented.

## Review

Two adversarial sub-agent reviews (correctness, simplicity) ran against the branch; all accepted findings are applied in the final commit — including a real bug where every API tagged error printed an empty CLI diagnostic (now human messages, with tightened black-box assertions that can no longer pass vacuously).

## Test plan

- `mise run app-server:check` — fmt, strict types, 80 tests, OpenAPI drift ✅
- `mise run contract:check` — + Swift client build/tests ✅
- `mise run -C app-server test:compiled` — compiled artifact: creates+migrates DB in a fresh data dir, restart is a clean no-op, flag-less CLI against the running server, 403 on cross-origin, clean SIGTERM ✅

### DIFF
--- a/app-server/src/api.ts
+++ b/app-server/src/api.ts
@@ -67,9 +67,12 @@ export const CreateProjectRequest = Schema.Struct({
   description: "Payload for creating a project.",
 })
 
+// optionalKey (absent key), not optional (key-or-undefined): the undefined
+// union renders as an anyOf-with-null the pinned Swift generator drops,
+// which would strip these fields from the generated client entirely.
 export const UpdateProjectRequest = Schema.Struct({
-  name: Schema.optional(ProjectName),
-  defaultWorkingDirectory: Schema.optional(
+  name: Schema.optionalKey(ProjectName),
+  defaultWorkingDirectory: Schema.optionalKey(
     AbsolutePath.annotate({
       description: "Existing directory; stored canonicalized (symlinks resolved).",
     }),
@@ -96,9 +99,13 @@ export const FsCheckResponse = Schema.Struct({
   }),
   state: DirectoryState,
   checkedAt: Schema.String.annotate({ description: "Check time (ISO 8601 UTC)." }),
-  reason: Schema.NullOr(Schema.String).annotate({
-    description: 'Why the state is not conclusive (e.g. "timeout" for unknown).',
-  }),
+  // Absent when the state is conclusive (not null — see UpdateProjectRequest).
+  reason: Schema.optionalKey(
+    Schema.String.annotate({
+      description:
+        'Why the state is not conclusive (e.g. "timeout" for unknown); absent otherwise.',
+    }),
+  ),
 }).annotate({
   identifier: "FsCheckResponse",
   description: "Result of a directory health check. Never persisted.",
--- a/app-server/src/api.ts
+++ b/app-server/src/api.ts
@@ -104,6 +104,10 @@ export const FsCheckResponse = Schema.Struct({
   description: "Result of a directory health check. Never persisted.",
 })
 
+// The error classes carry human `message`s so every consumer (the CLI above
+// all) can print a real diagnostic — a TaggedErrorClass message is otherwise
+// empty.
+
 /** Unknown project id. */
 export class ProjectNotFound extends Schema.TaggedErrorClass<ProjectNotFound>()(
   "ProjectNotFound",
@@ -113,7 +117,11 @@ export class ProjectNotFound extends Schema.TaggedErrorClass<ProjectNotFound>()(
     description: "No project exists with the given id.",
     httpApiStatus: 404,
   },
-) {}
+) {
+  override get message(): string {
+    return `no project with id ${this.projectId}`
+  }
+}
 
 /** The directory failed validation; `state` carries the tagged failure. */
 export class DirectoryUnavailable extends Schema.TaggedErrorClass<DirectoryUnavailable>()(
@@ -127,7 +135,18 @@ export class DirectoryUnavailable extends Schema.TaggedErrorClass<DirectoryUnava
     description: "The directory does not exist, is not a directory, or cannot be read/traversed.",
     httpApiStatus: 422,
   },
-) {}
+) {
+  override get message(): string {
+    switch (this.state) {
+      case "missing":
+        return `directory ${this.path} does not exist`
+      case "inaccessible":
+        return `directory ${this.path} cannot be read or traversed`
+      case "not_directory":
+        return `${this.path} is not a directory`
+    }
+  }
+}
 
 /** The bounded directory check did not complete; retryable, fail-closed. */
 export class DirectoryCheckTimedOut extends Schema.TaggedErrorClass<DirectoryCheckTimedOut>()(
@@ -139,7 +158,11 @@ export class DirectoryCheckTimedOut extends Schema.TaggedErrorClass<DirectoryChe
       "The directory check did not complete within its bounded timeout. Retryable — a timeout is not proof the path is missing.",
     httpApiStatus: 422,
   },
-) {}
+) {
+  override get message(): string {
+    return `the check of directory ${this.path} timed out; retry`
+  }
+}
 
 const projectIdParam = { projectId: Schema.String }
 
--- a/app-server/src/api.ts
+++ b/app-server/src/api.ts
@@ -30,6 +30,119 @@ export const VersionResponse = Schema.Struct({
   description: "Build and contract version information for the running server.",
 })
 
+/** A server-host absolute path. */
+export const AbsolutePath = Schema.String.check(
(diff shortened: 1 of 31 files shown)


app-server/src/api.ts
```diff
@@ -30,6 +30,142 @@ export const VersionResponse = Schema.Struct({
...
+  defaultWorkingDirectory: AbsolutePath.annotate({
+    description: "Existing directory; stored canonicalized (symlinks resolved).",
+  }),
+}).annotate({
+  identifier: "CreateProjectRequest",
+  description: "Payload for creating a project.",
+})
+
+export const UpdateProjectRequest = Schema.Struct({
+  name: Schema.optional(ProjectName),
```
jeremytondo: **[P1] Preserve PATCH fields in the generated Swift client**

`mise run contract:check` warns that the generated schema's `null` branch is unsupported, then emits `UpdateProjectRequest` as an empty Swift struct (`init()` and `knownKeys: []`). Consequently `ATCAppServerAPI` cannot send either update field; Swift PATCH callers can only encode `{}`. Please express omission with a schema the pinned generator preserves (an exact optional key appears to be the relevant Effect v4 distinction) and add a Swift regression test that constructs/encodes both fields.

app-server/src/api.ts
```diff
@@ -30,6 +30,142 @@ export const VersionResponse = Schema.Struct({
...
+  description: "Tagged result of a demand-driven directory health check.",
+})
+
+export const FsCheckResponse = Schema.Struct({
+  path: Schema.String.annotate({
+    description: "The checked path (canonicalized when the directory is available).",
+  }),
+  state: DirectoryState,
+  checkedAt: Schema.String.annotate({ description: "Check time (ISO 8601 UTC)." }),
+  reason: Schema.NullOr(Schema.String).annotate({
```
jeremytondo: **[P1] Keep `reason` decodable by the Swift client**

The pinned Swift generator also warns on this `null` branch and drops `reason` entirely from generated `FsCheckResponse`, including its strict `knownKeys` set. The handler always returns `reason` (`null` for normal states and a string on timeout), so every valid `/fs/check` payload is then rejected as containing an undocumented property. Please use a generator-compatible representation and add a Swift decoding test for both null and non-null cases.

app-server/src/config.ts
```diff
@@ -0,0 +1,215 @@
...
+    if (!settings.dataDir.startsWith("/")) {
+      return yield* Effect.fail(
+        new ConfigLoadError({
+          source: `${configFile} / ATC_DATA_DIR`,
+          message: `dataDir must be an absolute path, got "${settings.dataDir}"`,
+        }),
+      )
+    }
+
+    const stateDir = xdgDir(env, "XDG_STATE_HOME", [".local", "state"])
```
jeremytondo: **[P2] Reject relative state paths too**

`XDG_STATE_HOME=relative-state` currently settles to `stateDir === "relative-state/atc"`, so `Logging.layer` writes `atc.log` relative to the launch directory. That breaks this PR's invariant that compiled behavior never depends on the cwd; config and data locations already reject the equivalent input. Please apply the same absolute-path validation to the state location (ideally through shared path validation) and cover it in `config.test.ts`.

jeremytondo (commented):
I found three blocking issues: two make the generated Swift client unusable for parts of this contract despite `contract:check` succeeding, and one violates the configuration invariant that compiled behavior never depends on the working directory. The app-server suite (80 enabled tests), cross-language contract build/tests, and both GitHub workflows otherwise pass.

app-server/src/api.ts
jeremytondo: Fixed in 7193f99. `Schema.optional` is `optionalKey(UndefinedOr(...))` — the undefined union rendered as the anyOf-with-null the generator drops. Both fields now use `Schema.optionalKey` (the exact-optional form), so the generated `UpdateProjectRequest` carries both properties again, and a Swift regression test (`updateProjectFields` in ClientTests.swift) encodes both and asserts the exact wire JSON. The handler and empty-patch semantics are unchanged; the CLI now builds the payload with spread-conditionals so omitted flags are truly absent keys.

app-server/src/api.ts
jeremytondo: Fixed in 7193f99. `reason` is now `Schema.optionalKey(Schema.String)` — absent when the state is conclusive, present (`"timeout"`) on the unknown path — instead of a required nullable field. The generated `FsCheckResponse` keeps `reason` in its `knownKeys`, and `fsCheckReason` in ClientTests.swift decodes both the with-reason and without-reason payloads. `directories.ts`, the handler, and the TS tests were updated to the absent-key shape (this also un-lies the `required` list in the document).

app-server/src/config.ts
jeremytondo: Fixed in 7193f99. Extracted a shared `requireAbsolute` helper in `config.ts`; the config file path, `dataDir`, and now the state directory all go through it, so `XDG_STATE_HOME=relative-state` fails fast naming the source. Covered by a new `config.test.ts` case.

PR #128, opened 2026-08-02
ATC-129: zmx TerminalAdapter, Subprocess PTY extension, and fake-adapter seam (PR 1 of 2)

jeremytondo:
The structural core of zmx-backed Terminal support ([ATC-129](https://linear.app/elevenideas/issue/ATC-129), part of [ATC-122](https://linear.app/elevenideas/issue/ATC-122)) — stacked PR 1 of 2. PR 2 (Terminal resources, WebSocket attach transport, CLI, reconciliation) builds on this branch.

## What's here

- **`Subprocess` PTY extension** (`src/subprocess.ts`): `spawnPty` on the existing seam using Bun's native PTY (`Bun.spawn` + `terminal:`), with the same scoped kill+reap guarantees (SIGTERM → SIGKILL escalation, PTY released on scope close). Bounded sliding output queue, decoded `outputTail` for diagnostics, short-write detection.
- **`TerminalAdapter` seam** (`src/terminalAdapter.ts`): the narrow interface the rest of the app talks to — derived session names (`atc-` + terminal id's 32 hex chars, reversible, never persisted), `ZmxUnavailable` (retryable, "inventory says nothing") vs `SessionOperationFailed` (conclusive) vs `SessionNotFound` error vocabulary.
- **zmx implementation** (`src/zmxAdapter.ts`): private `ZMX_DIR` under the state dir (0700), mandatory env scrubbing (`ZMX_SESSION`, `ZMX_SESSION_PREFIX`), socket-path length validated at boot (103/107 bytes by platform), explicit executable resolution with one actionable diagnostic, tolerant `zmx list` parsing where `err=` entries are existing-but-unreachable.
- **Config** (`src/config.ts`): `zmxExecutable` key (file / `ATC_ZMX_EXECUTABLE`, default `zmx` on PATH; bare-name-or-absolute validated) and `terminalSocketDir`.
- **Tests**: deterministic adapter suite against a `fake-zmx` fixture (per-test wrapper scripts, no `process.env` mutation), PTY seam tests, fake in-memory adapter for PR 2 consumers with honesty tests, and opt-in real-zmx smoke tests (`mise run -C app-server test:zmx`, verified green against zmx 0.6.0).

## Behavioral invariants encoded (from the ATC-122 decision record)

- Only a **complete inventory** proves absence; only a **reachable** entry proves a live session. Exit codes, PTY EOF, and attach failures never mark anything ended.
- `zmx attach` auto-creates, so: create pre-flights (creating is never silently attaching — a name collision fails conclusively), attach pre-flights and then verifies by daemon pid that it reached the original session, killing a silently resurrected phantom.
- `zmx kill` returns before death: absence is verified by polling complete inventory passes. Verification is **pass-counted, not wall-clock**, so a slow inventory extends the wait instead of manufacturing a conclusive failure; each zmx command has its own timeout mapping to retryable `ZmxUnavailable`.
- A command that finishes before its session settles is a failed launch (with a captured output tail as the diagnostic).

## Boundary note for reviewers

`ZmxUnavailable` is defined in `terminalAdapter.ts` here; PR 2 moves it into the contract (`api.ts`, 503) because the API surfaces it directly. Likewise the temp-dir test helpers and the `test:zmx` task grow in PR 2 (recovery suite, compiled-binary dependency). Reviewed in isolation, those placements are intentional way-points, not final homes.

## Review process

Implemented, then adversarially reviewed by two independent reviewers (correctness, simplicity). The correctness review reproduced a critical bug against real zmx (an unreachable `err=` inventory entry satisfied the create settle-poll) which is fixed here with the reachable-qualified predicate plus a regression test; also fixed: create-over-existing silently attaching, attach TOCTOU resurrection, latency-as-failure in verification, bad-cwd misclassified as retryable, world-readable socket dir, Linux socket-path limit. The simplicity review drove the spec dedup (`PtySpawnSpec extends SpawnSpec`), the shared `outputTail`, wrapper-baked fixture config, and shared test helpers.

Known coverage gap: the attach TOCTOU resurrection guard has no deterministic unit test (the race window can't be scripted with the fixture); it is exercised structurally and against real zmx in the smoke suite.

### DIFF
--- a/app-server/src/subprocess.ts
+++ b/app-server/src/subprocess.ts
@@ -1,5 +1,15 @@
-import { Cause, Context, Duration, Effect, Layer, Queue, Schema, Scope, Stream } from "effect"
-import type { PlatformError } from "effect/PlatformError"
+import {
+  Cause,
+  Context,
+  Deferred,
+  Duration,
+  Effect,
+  Layer,
+  Queue,
+  Schema,
+  Scope,
+  Stream,
+} from "effect"
 import { ChildProcess, ChildProcessSpawner } from "effect/unstable/process"
 
 // The generic subprocess seam: how the App Server launches and supervises
@@ -12,10 +22,18 @@ import { ChildProcess, ChildProcessSpawner } from "effect/unstable/process"
 const STDERR_TAIL_LINES = 100
 /** Individual captured lines are truncated to keep diagnostics bounded. */
 const MAX_CAPTURED_LINE_LENGTH = 2_000
+/** How much recent PTY output (UTF-16 units) to keep for diagnostics. */
+const PTY_OUTPUT_TAIL_LENGTH = 2_000
+/**
+ * PTY output chunks buffered ahead of the consumer. Terminals are lossy on
+ * overflow by nature (the multiplexer repaints on reattach), so the oldest
+ * chunks slide out rather than growing memory without bound.
+ */
+const PTY_OUTPUT_QUEUE_CAPACITY = 1_024
 
 export class SubprocessError extends Schema.TaggedErrorClass<SubprocessError>()("SubprocessError", {
   executable: Schema.String,
-  operation: Schema.Literals(["spawn", "stdin", "stdout", "exit"]),
+  operation: Schema.Literals(["spawn", "stdin", "stdout", "exit", "resize"]),
   message: Schema.String,
 }) {}
 
@@ -31,6 +49,16 @@ export interface SpawnSpec {
   /** Grace period before SIGKILL when the child ignores SIGTERM at cleanup. */
   readonly forceKillAfter?: Duration.Input
 }
+
+export interface PtySpawnSpec extends SpawnSpec {
+  /** Initial pseudo-terminal size. */
+  readonly cols?: number
+  readonly rows?: number
+}
+
+/** Default PTY size when the caller has no client-provided dimensions yet. */
+export const DEFAULT_PTY_COLS = 80
+export const DEFAULT_PTY_ROWS = 24
 
 export interface Child {
   readonly pid: number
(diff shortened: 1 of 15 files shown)


app-server/src/subprocess.ts
```diff
@@ -170,9 +236,111 @@ const spawn = Effect.fnUntraced(function* (
...
+    Effect.try({ try: run, catch: subprocessError(spec.executable, operation) })
+
+  const child: PtyChild = {
+    pid: proc.pid,
+    output: Stream.fromQueue(output),
+    outputTail: Effect.sync(() => outputTail),
+    write: (data) =>
+      terminalOp("stdin", () => {
+        const length = typeof data === "string" ? Buffer.byteLength(data) : data.byteLength
+        const written = terminal?.write(data) ?? 0
```
jeremytondo, line 337: [P1] Track PTY exit state instead of trusting `terminal.write` here. Both Ubuntu jobs in App Server CI fail `fails writes after the terminal closes`: after `yield* child.exitCode`, Bun on Linux continues returning the full byte count for at least 1s, so this method reports success while discarding input. That violates the `PtyChild.write` contract and leaves Linux callers unable to detect a dead attach client. Please record closure from the PTY `exit` callback / `proc.exited` and fail writes once closed (while retaining the short-write check for live processes).

app-server/src/zmxAdapter.ts
```diff
@@ -0,0 +1,358 @@
...
+        return yield* Effect.fail(
+          new ZmxConfigError({
+            reason:
+              `terminal socket directory ${socketDir} is too deep: socket paths would exceed ` +
+              `${MAX_SOCKET_PATH_BYTES} bytes; move the state directory (XDG_STATE_HOME)`,
+          }),
+        )
+      }
+      // Private: sockets grant full read/write on the user's terminals.
+      yield* fs.makeDirectory(socketDir, { recursive: true, mode: 0o700 })
```
jeremytondo, line 138: [P2] `mode: 0o700` only affects a directory when it is created; `recursive` mkdir leaves an existing `terminalSocketDir` at its prior mode. If this path was pre-created (for example by a prior/manual zmx invocation) as 0750/0755/0777, ATC proceeds while claiming the socket directory is private even though other users may be able to reach terminal sockets. Please explicitly `chmod(socketDir, 0o700)` after ensuring it exists, and cover the pre-existing permissive-directory case.

jeremytondo (commented):
Two blocking reliability/security issues found. The Linux-specific PTY lifecycle issue is already failing both Ubuntu App Server CI jobs; the final stacked code passes the macOS App Server and contract gates locally.

app-server/src/subprocess.ts
jeremytondo: Fixed in b8425f4: closure is now tracked explicitly (`ptyClosed`, set from both the PTY `exit` callback and `proc.exited`, registered before `exitCode` can resolve), and `write` fails with a tagged `stdin` error once closed while retaining the short-write check for live processes. The existing `fails writes after the terminal closes` test covers the Linux behavior.

app-server/src/zmxAdapter.ts
jeremytondo: Fixed in b8425f4: boot now explicitly `chmod`s the socket directory to 0700 after ensuring it exists, so a pre-existing permissive directory is re-tightened rather than trusted. Covered by a new adapter test that pre-creates the directory at 0755 and asserts 0700 after layer boot.

PR #129, opened 2026-08-02
ATC-130: Terminal resources, WebSocket attach transport, CLI, and reconciliation (PR 2 of 2)

jeremytondo:
Terminal resources and transport on top of the ATC-129 adapter core ([ATC-130](https://linear.app/elevenideas/issue/ATC-130), part of [ATC-122](https://linear.app/elevenideas/issue/ATC-122)) — stacked PR 2 of 2, based on #128. **Merge #128 first, then retarget/merge this one.**

## What's here

- **Model + persistence**: migration `0002_terminals` (STRICT, FK `ON DELETE RESTRICT`), `TerminalRepository` (UUIDv7 id, immutable JSON command argv, canonicalized initial working directory, mutable label, `starting`/`live`/`ended` with idempotent tombstoning — the first observation of death wins).
- **Domain service** (`src/terminals.ts`): two-phase create (`starting` row → zmx session launch → `live`; any non-success exit including a client-abort interruption kills the session best-effort and leaves no record), demand-driven reconciliation (startup during layer build resolves live/starting rows and cleans orphan sessions in the private dir; request-time reconcile on list/read/attach only tombstones live rows — rows are snapshotted *before* the inventory so a concurrently completing create can never be falsely tombstoned), delete = kill → verify absence → remove (unverifiable kills fail retryably), `confirmEnded` as the bridge's authoritative-absence check.
- **Contract** (`api.ts` + regenerated `openapi.json`): Terminal CRUD with pinned operation ids, errors `TerminalNotFound` 404 / `ZmxUnavailable` 503 (retryable) / `TerminalLaunchFailed` 422 / `ProjectHasTerminals` 409 (project deletion RESTRICTed while terminals exist), and the WebSocket attach endpoint declared in the contract but `OpenApi.Exclude`d from the document (the Swift generator consumes the artifact directly). Swift client builds and passes against the new contract.
- **WS attach bridge** (`src/terminalAttach.ts`): pre-upgrade rejection with real HTTP statuses (404 / 410 tombstone / 426 no-upgrade), binary frames = bytes, text frames = JSON control (`resize`, app-level `ping`/`pong` — Bun exposes no transport ping), close vocabulary stated once on the contract endpoint (1000 `terminal_ended` only after confirmed absence; 1011 `attach_failed`/`zmx_unavailable`/`ping_timeout` retryable). Bridges are forked into the handler layer's scope because Bun aborts the upgraded request's fiber (found empirically); server shutdown reaps every bridge.
- **CLI**: `atc terminal list|get|create|rename|delete` via the contract client (create takes a trailing exec argv), and `atc terminal attach` in raw mode (Ctrl-] detaches) — pre-flighted over the typed API so unknown/ended terminals get real diagnostics instead of a generic connect error.
- **Tests**: service suite over the fake adapter (including startup recovery: adopt/tombstone/drop/orphan-clean), black-box WS bridge suite against a spawned serve with `ATC_ZMX_EXECUTABLE` pointed at the fake-zmx wrapper (in-process WS upgrades through the Bun adapter don't complete under vitest — and this exercises the real adapter + PTY + bridge end to end), CLI black-box additions including an attach e2e, and an opt-in compiled restart-recovery suite against real zmx (`mise run -C app-server test:zmx`): terminals survive a server restart, orphans are cleaned, attach works against a recovered session.

## Notable empirical findings encoded here

- Bun never completes a graceful `server.stop()` after a **server-initiated** WS close; the server layer now bounds graceful shutdown at 2s (comment in `server.ts`).
- Bun aborts the upgraded request's fiber (`ClientAbort`), so the bridge cannot live in the request effect.

## Review process

Implemented, then adversarially reviewed by two independent reviewers (correctness, simplicity); all majors fixed: reconcile row/inventory ordering (falsely-tombstoned concurrent create, reproduced), interrupted-create cleanup (invisible `starting` rows blocking project deletion, reproduced end-to-end), CLI attach diagnostics, plus ~30 smaller items (OpenApi.Exclude instead of a hand-rolled stripper, repository signature honesty, shared test helpers, temp-dir tracking). Known accepted gaps: `deleteProject`'s count-then-delete is not transactional (a racing create falls back to the FK RESTRICT defect — single-user app), and the bounded outbox cannot apply true send-backpressure because Bun's WS writer exposes none (memory stays bounded via the sliding PTY buffer; dead clients reaped by ping timeout).

### DIFF
--- a/app-server/src/terminalAttach.ts
+++ b/app-server/src/terminalAttach.ts
@@ -3,13 +3,24 @@ import type { Scope } from "effect"
 import type { HttpServerRequest } from "effect/unstable/http"
 import { HttpServerResponse } from "effect/unstable/http"
 import * as Socket from "effect/unstable/socket/Socket"
+import {
+  clampDimension,
+  CLOSE_ATTACH_FAILED,
+  CLOSE_PING_TIMEOUT,
+  CLOSE_TERMINAL_ENDED,
+  CLOSE_ZMX_UNAVAILABLE,
+  decodeControlFrame,
+  encodeControlFrame,
+} from "./attachProtocol.ts"
 import { DEFAULT_PTY_COLS, DEFAULT_PTY_ROWS } from "./subprocess.ts"
+import type { SessionSize } from "./terminalAdapter.ts"
 import type { Terminals } from "./terminals.ts"
 
 // The WebSocket attach bridge (ATC-130): one upgraded socket bridged onto
-// one scoped `zmx attach` PTY client. Wire protocol and close vocabulary
-// are stated once, on the contract endpoint (api.ts `attachTerminal`);
-// this module implements them.
+// one scoped `zmx attach` PTY client. The wire protocol and close
+// vocabulary live in attachProtocol.ts (documented in prose on the contract
+// endpoint, api.ts `attachTerminal`); this module implements the server
+// side.
 //
 // Flow control: Bun's WS writer exposes no send-backpressure signal, so a
 // slow client cannot throttle the PTY. Memory stays bounded anyway — the
@@ -23,34 +34,6 @@ const PING_INTERVAL_MILLIS = 30_000
 const PONG_TIMEOUT_MILLIS = 120_000
 /** Outbound frames buffered ahead of the socket writer. */
 const OUTBOX_CAPACITY = 256
-
-const clampDimension = (raw: string | number | undefined, fallback: number): number => {
-  const parsed = typeof raw === "number" ? raw : Number.parseInt(raw ?? "", 10)
-  return Number.isNaN(parsed) ? fallback : Math.max(1, Math.min(1_000, Math.floor(parsed)))
-}
-
-const decodeControl = (
-  text: string,
-): { type: "resize"; cols: number; rows: number } | { type: "pong" } | { type: "ignored" } => {
-  try {
-    const parsed = JSON.parse(text) as { type?: string; cols?: number; rows?: number }
-    if (
-      parsed.type === "resize" &&
-      typeof parsed.cols === "number" &&
-      typeof parsed.rows === "number"
-    ) {
-      return {
-        type: "resize",
-        cols: clampDimension(parsed.cols, DEFAULT_PTY_COLS),
-        rows: clampDimension(parsed.rows, DEFAULT_PTY_ROWS),
-      }
-    }
-    if (parsed.type === "pong") return { type: "pong" }
-  } catch {
-    // Unknown or malformed control frames are ignored, never fatal.
-  }
-  return { type: "ignored" }
-}
 
 /**
  * The raw handler implementing the contract's `attachTerminal` endpoint.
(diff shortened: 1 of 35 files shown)


app-server/src/terminals.ts
```diff
@@ -0,0 +1,275 @@
...
+        const sessionName = sessionNameForTerminalId(record.id)
+        switch (record.status) {
+          case "live":
+            if (!present.has(sessionName)) yield* repository.markEnded(record.id)
+            else claimed.add(sessionName)
+            break
+          case "starting":
+            // Crash mid-create: adopt the session if it made it, otherwise
+            // the failed launch leaves no record.
+            if (present.has(sessionName)) {
```
jeremytondo: [P1] Do not promote an unreachable session to `live`. `present` includes every inventory entry, including `reachable: false`, so a crash that leaves a wedged socket turns its `starting` row into a public live terminal at restart. This contradicts ATC-122’s core invariant that only a reachable entry proves liveness (and differs from `createSession`, which deliberately waits for `reachable`). Please branch on the matching `SessionInfo`: adopt only when reachable; an unreachable match must remain unresolved/retryable and must be claimed so orphan cleanup does not kill it. Add a startup-reconciliation regression test with an unreachable matching session.

app-server/src/terminalAttach.ts
```diff
@@ -0,0 +1,219 @@
...
+    const readSocket = socket.runRaw((message) => {
+      if (typeof message === "string") {
+        const control = decodeControl(message)
+        if (control.type === "resize") {
+          return Effect.ignore(connection.resize({ cols: control.cols, rows: control.rows }))
+        }
+        if (control.type === "pong") lastPong = Date.now()
+        return undefined
+      }
+      return Effect.ignore(connection.write(message))
```
jeremytondo: [P2] Propagate PTY input failures into the bridge’s retryable close path instead of suppressing them. If `connection.write` fails (closed PTY, short write, or the corrected Linux post-exit guard from #128), `Effect.ignore` silently drops the user’s input and leaves the WebSocket open until some unrelated EOF/timeout occurs. The wire contract already defines 1011 `attach_failed`; please terminate the bridge with that reason on write failure, without inferring that the terminal ended.

jeremytondo (commented):
Two blocking lifecycle/transport issues found. The final stacked implementation passes the local App Server gate (135 tests plus format/type/OpenAPI checks) and the end-to-end contract gate (including 57 Swift tests), but PR #129 currently has no GitHub Actions run.

app-server/src/terminals.ts
jeremytondo: Fixed in f418ac6: startup reconciliation now branches on reachability for `starting` rows — only a reachable inventory entry adopts the row as live; an unreachable match leaves the row `starting` (hidden, resolved at a later startup) and claims the session so orphan cleanup spares it; absence still deletes the row. Regression test added to the startup-reconciliation suite (a `wedged` starting row with an unreachable session).

app-server/src/terminalAttach.ts
jeremytondo: Fixed in f418ac6: a failed `connection.write` now enqueues `1011 attach_failed` onto the outbox, terminating the bridge through the ordered close path instead of silently dropping input — without consulting the inventory, so `terminal_ended` is never inferred. Covered by a new in-process bridge test (`terminalBridge.test.ts`) driving `runBridge` over an in-memory socket with a fake adapter whose PTY writes fail.

PR #134, opened 2026-08-05
ATC-138: Reorganize the App Server into domain folders with a top-level composition root

jeremytondo:
Reorganizes `app-server/src` from 29 flat files into one-level domain folders with surfaces (`cli/`, `api/`) separated from core domains (`projects/`, `terminals/`, `agents/`), shared infrastructure in `platform/`, and one top-level composition root (`server.ts`). Lands the structure before Threads (ATC-124) begins so that work has an obvious home.

Ordered commits, each standalone-green on `mise run -C app-server check` and `mise run contract:check`:

1. **Extract `Projects` domain service** — canonicalize-on-create/update and the delete-restricted-while-terminals-exist guard move verbatim out of `handlers.ts`; handlers become pure delegation for projects, as they already were for terminals.
2. **Move the composition root into `server.ts`** — the production layer assembly leaves the serve command; `server.ts` now exports both the open `routes`/`layer` (tests compose with fakes) and the closed `production` assembly. The CLI keeps zero domain-layer imports.
3. **Split `cli.ts` into `cli/`** — `cli.ts` (shared plumbing: diagnostics, base-URL seam, `clientCommand`, plus health/version/fs), `serve.ts`, `projects.ts`, `terminals.ts`, `gateway.ts`. The root `atc` assembly moves to `main.ts` per the issue's target layout (also avoids an ESM eval-order cycle between the plumbing module and the command modules).
4. **Pure move into domain folders** — `api.ts` becomes `api/contract.ts`; everything else keeps its name. Tests mirror the structure (`test/api/`, `test/platform/`, `test/cli/`, `test/terminals/`, `test/agents/`) with the shared harnesses, `server.test.ts`, `compiled.blackbox.test.ts`, and `fixtures/` at the test root. Opt-in mise task paths updated in the same commit so `test:zmx`/`test:smoke` stay correct.
5. **Path-reference sweep** — AGENTS.md (contract path, CLI seam wording), READMEs, module comments.
6. **Review feedback** — `Projects` reads `TerminalRepository.countForProject` directly (mirrors `Terminals` depending on `ProjectRepository` directly; keeps zmx/startup-reconciliation out of the Projects dependency graph); the now-unused `Terminals.countForProject` pass-through is removed.
7. **Stale references found in review** — `packages/contracts/README.md`, two module headers, and the AGENTS.md canonical-name rule updated for `cli/` command modules.

**Nothing public changes**: `openapi.json` is byte-identical (absent from the diff), all pre-existing service tags and operation ids unchanged; the only new tag is `app-server/Projects`. Verified by adversarial review including a live diff of both revisions against fake-zmx (identical HTTP behavior across create/update/delete edge cases) and a full `--help`-output diff of every CLI command (identical).

Linear: [ATC-138](https://linear.app/elevenideas/issue/ATC-138/reorganize-the-app-server-into-domain-folders-with-a-top-level)

https://claude.ai/code/session_018Rd6RAK5eJMvN3EYfJyr17


jeremytondo (commented):
Completed a full review of the seven-commit series. I checked the `Projects` service extraction against the pre-PR handler behavior, the production Layer composition, the CLI split/root command assembly, test/fixture path moves, stale path references, and OpenAPI stability. I found no actionable issues.

Verification:
- `TMPDIR=/tmp mise run -C app-server check` — 195 passed, 9 skipped
- `TMPDIR=/tmp mise run contract:check` — passed, including 57 Swift tests
- OpenAPI artifact is unchanged
- All five GitHub checks are green

PR #136, opened 2026-08-06
ATC-140: Agents registry and the AgentAdapter seam extensions

jeremytondo:
PR 2 of 3 for [ATC-124](https://linear.app/elevenideas/issue/ATC-124). Implements [ATC-140](https://linear.app/elevenideas/issue/ATC-140). **Stacked on #135** (base branch is the ATC-139 branch; merge #135 first, then retarget or merge this).

## What this adds

- **Seam extensions on `AgentAdapter`** — the uniform contracts behind "provider sessions materialize on first interaction":
  - `prepareTuiSession({ cwd }) → { launchSpec, identity }`: one fresh-launch shape for every provider. Claude pre-assigns a minted UUID (`--session-id`, probe-verified) so `identity` resolves immediately; Codex captures the launched TUI's `thread/started` broadcast on the held observer connection, cwd-checked fail-closed, with a per-provider launch lock held for the caller's scope so the correlation is deterministic. **The lock also serializes Codex `createSession`'s own `thread/start`** — the provider echoes `thread/started` for every thread, so an ATC-driven create in the same cwd could otherwise be mis-adopted as the TUI's identity (adversarial-review catch, reproduced live before fixing).
  - `observeSession` — the one normalized per-session activity subscription for TUI-driven sessions (Claude: webhook-fed; Codex: shared-server status fan-out that survives reconnects).
  - `checkSession` — demand-driven reconciliation (Codex: `thread/list`; Claude: honestly `unknown` — no cheap truthful provider query exists, so linked-terminal liveness carries staleness re-derivation in V1).
  - `releaseSession` — launch-file/secret cleanup for thread deletion.
  - `tuiLaunch` now takes the thread's opaque `providerMetadata` and returns `{ launchSpec, providerMetadata? }` so adapter state minted at launch (the Claude hook secret) can always be persisted.
- **The opaque `provider_metadata` path**: the Claude hook secret is minted once, rides the thread's metadata, and is adopted back after ATC restarts (`ClaudeHooks.adoptSecret`/`revokeSecret`), keeping a running TUI's hooks valid across restarts. Abandoned prepares clean up their secret registration and 0600 hook files.
- **Read-only Agents registry**: `GET /agents`, `GET /agents/:agentId` (`AgentNotFound` 404) with availability + detected version (demand-driven, bounded, never persisted) and the seam's capability report; `AgentRegistry.adapterFor` is the domain's only door to providers. No CLI commands — `atc api` covers it.
- **Production wiring**: the provider adapters join the server assembly (previously smoke-only). Verified: boot spawns no provider process; detection runs only on `/agents` reads.

## Review hardening folded in

Two adversarial reviews (correctness with live probes, simplicity). Beyond the mis-capture fix above: the launch-lock acquisition is interruptible (a queued open stays cancellable), an abandoned capture completes its identity gate instead of hanging awaiters, `readInstalledVersion` is bounded (a wedged binary can't hang the request), the fakes got per-slug capability reports so a swapped registry mapping can't pass tests, and speculative test-harness surface was cut.

## Tests

Per-adapter mechanics: Claude pre-assignment + hook plumbing + secret restore/revoke against hook fixtures; Codex capture (incl. foreign-cwd rejection), passive observation, and `thread/list` reconciliation against the fake app-server listener. Registry behavior + OpenAPI document/runtime cases. `mise run check` (217 tests) and `mise run contract:check` green.

### DIFF
--- a/app-server/src/agents/claudeAdapter.ts
+++ b/app-server/src/agents/claudeAdapter.ts
@@ -59,8 +59,9 @@ import * as Subprocess from "../platform/subprocess.ts"
 //     NOT bridged into these feeds: a TUI-driven session has no adapter
 //     connection by definition, and while ATC drives, the same vocabulary
 //     already arrives as in-process SDK callbacks — a webhook delivery for
-//     a live connection could only be stale or spoofed. TUI-side consumers
-//     subscribe to claudeHooks directly (ATC-124).
+//     a live connection could only be stale or spoofed. TUI-driven activity
+//     flows through observeSession below, which is the claudeHooks
+//     subscriber; consumers stay above the seam.
 
 /** The Claude Code version this adapter was validated against. */
 export const CLAUDE_TESTED_VERSION = "2.1.221"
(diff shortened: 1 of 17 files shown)


app-server/src/agents/codexAdapter.ts
```diff
@@ -60,6 +66,12 @@ const ThreadReply = Schema.Struct({
 
 const TurnReply = Schema.Struct({ turn: Schema.Struct({ id: Schema.String }) })
 
+const ThreadListReply = Schema.Struct({
+  threads: Schema.Array(
```
jeremytondo: [P1] Decode and paginate the real `thread/list` response. The checked-in live probe reads `result.data` / `nextCursor`, and the pinned generated Codex schema defines `V2ThreadListResponse` as `{ data, nextCursor }`; this decoder expects `{ threads }`, while the fake listener was changed to emit that same incorrect shape. As a result, every real `checkSession` call fails decoding and is mapped to `AgentUnavailable`, so demand-driven stale-state reconciliation never works for Codex. Please model the actual paginated response, walk pages until the target is found or the cursor ends, and update the fake/test to pin the real wire shape.

app-server/src/agents/claudeAdapter.ts
```diff
@@ -411,6 +412,60 @@ export const layerWith = (adapterOptions: ClaudeAdapterOptions) =>
...
+       * never in any argv (ps would show it, including curl's) and never in
+       * the URL (request paths land in tracer span names). Files are
+       * overwritten per launch and removed by releaseSession.
+       */
+      const ensureHookPlumbing = (providerSessionId: string, metadata: string | undefined) =>
+        Effect.gen(function* () {
+          const known = metadataSecret(metadata)
+          const secret = known ?? (yield* hooks.registerSecret(providerSessionId))
+          if (known !== null) yield* hooks.adoptSecret(providerSessionId, known)
+          const url = `http://127.0.0.1:${config.port}/internal/claude/hooks`
```
jeremytondo, line 452: [P2] Install cleanup before hook-plumbing allocation can fail. `ensureHookPlumbing` registers a freshly minted secret before these file operations, but `prepareTuiSession` does not register its cleanup finalizer until this helper returns. If mkdir/write/chmod (especially the second file) fails, the effect exits with the secret still registered and may leave a partial header/settings file for a UUID that is never persisted. Please bracket the registration/files as one allocation or clean them on error, and cover a write failure after secret registration.

jeremytondo (commented):
Two actionable provider-seam issues found; both affect production behavior that the current fakes do not exercise.

app-server/src/agents/codexAdapter.ts
jeremytondo: Fixed in cb7c1e9. `ThreadListReply` now models the real paginated shape (`{ data, nextCursor }`), and `checkSession` walks pages until the target thread appears or the cursor runs out (with a repeated-cursor guard so a non-advancing cursor can't loop). The fake listener now serves the real wire shape one thread per page, and the test covers both a target on a later page and a full-walk miss.

app-server/src/agents/claudeAdapter.ts
jeremytondo: Fixed in cb7c1e9. The write sequence inside `ensureHookPlumbing` now cleans up on any failure or interruption (`Effect.onError`): partial files are removed, and the secret is revoked iff this call minted it — a metadata-carried secret may still be validating a running TUI, so those are never revoked on a failed relaunch. `prepareTuiSession` additionally brackets the plumbing with `Effect.acquireRelease`, so interruption can no longer land between the allocation and its finalizer. Two regression tests cover a failure before the first write and a failure on the second write (partial header removed, minted secret revoked — asserted via a spy on `registerSecret`).

PR #137, opened 2026-08-06
ATC-141: the openTerminal workflow — identity establishment, status wiring, thread open, recovery

jeremytondo:
PR 3 of 3 for [ATC-124](https://linear.app/elevenideas/issue/ATC-124). Implements [ATC-141](https://linear.app/elevenideas/issue/ATC-141). **Stacked on #136** (which is stacked on #135) — merge in order 135 → 136 → 137.

## What this adds

- **`threads.openTerminal`** (`POST /threads/:threadId/terminal` → Terminal, plus `atc thread open` = open + raw-TTY attach):
  - **Idempotent**: a live linked terminal is returned as-is — including one whose identity is still being established; at most one live linked TUI terminal per thread.
  - **Unconfirmed threads materialize fresh** through the uniform seam flow: `prepareTuiSession` → Terminal from `launchSpec` (nested-session env markers scrubbed uniformly via a new per-session env overlay on the terminal seam) → bounded identity await → persist session id + opaque metadata. The bound covers the whole flow (launch-lock queueing included), so a slow provider can't hang callers. On failure, defect, **or interruption** (a dropped client), the launched terminal is torn down — nothing survives that the thread adopted nothing from. Superseded unconfirmed sessions are unobserved + released before a successor is minted.
  - **Confirmed threads reopen their exact session** via `tuiLaunch`; mismatches fail closed; a delete racing the open adopts nothing.
  - `materialize` is the single identity-establishment transition, with openTerminal as its first caller (a native first-prompt is the future second caller — same persistence, no schema change).
- **Status wiring**: per-session activity subscriptions (keyed by thread **and** session, so a superseded session's feed can never drive — or falsely confirm — its successor) feed the in-memory `activityState`; the confirmed marker is set on the feed's busy→idle transition (a completed first turn — the same signal for every provider). Stale busy states with no live driver re-derive demand-driven via `checkSession`; a read after an ATC restart re-links and re-observes with no poller. Archive is refused while busy (`needs_input` counts — an approval prompt is still mid-turn).
- **Error mapping** per the decision record: `ProviderUnavailable` 503 (retryable), `ProviderSessionConflict` 409 (fail-closed), `ThreadArchived` 409.
- **Gated live smoke** (`test:smoke`): a real interactive `claude --session-id <minted> --settings <hooks>` launch confirms the pre-assigned id through its own hook delivery — the one implementation-time check the decision record left open, now verified passing on a real install.

## Review hardening folded in

Two adversarial reviews (correctness with reproduction probes, simplicity). Fixed from them: the superseded-session observation leak (old feed could confirm the new session and permanently wedge the thread), interruption-safe terminal cleanup (`onExit`, not `tapError`), release of superseded sessions' adapter resources, thread listing restored to one reconciled inventory pass per page (was N+1 zmx spawns), the idempotent open path no longer double-lists, reopen re-checks the row before inserting, and the open bound now covers launch-lock queueing. Regression tests cover each.

## Tests

`threadOpen.test.ts`: materialization/persistence/idempotency (env scrubbing included), confirmation + exact reopen, provisional re-materialization with feed-isolation and release assertions, timeout and interruption cleanup, archived blocking, release-on-delete, staleness re-derivation, and an in-process restart-recovery case (fresh process over the same DB re-observes on first read). CLI blackbox: `thread open` produces the one-line actionable 503 diagnostic when the provider is missing. `mise run check` (228 tests) and `mise run contract:check` green; `ATC_SMOKE=1` TUI smoke verified live.

### DIFF
--- a/app-server/src/threads/threads.ts
+++ b/app-server/src/threads/threads.ts
@@ -1,17 +1,38 @@
-import { Context, Effect, Layer } from "effect"
-import type { AgentActivity } from "../agents/agentAdapter.ts"
-import { Thread as ThreadSchema, ThreadBusy } from "../api/contract.ts"
+import { Context, Duration, Effect, Exit, Layer, Option, Scope, Stream } from "effect"
+import { AgentRegistry } from "../agents/agentRegistry.ts"
 import type {
+  AgentActivity,
+  AgentAdapter,
+  AgentConflict,
+  AgentIdentityMismatch,
+  AgentProtocolError,
+  AgentResumeFailed,
+  AgentUnavailable,
+  TuiLaunchSpec,
+} from "../agents/agentAdapter.ts"
+import { NESTED_SESSION_ENV_VARIABLES } from "../agents/agentAdapter.ts"
+import {
+  AGENT_IDS,
+  ProviderSessionConflict,
+  ProviderUnavailable,
+  Thread as ThreadSchema,
+  ThreadArchived,
+  ThreadBusy,
+  ThreadNotFound,
+} from "../api/contract.ts"
+import type {
+  AgentId,
   CreateThreadRequest,
   DirectoryCheckTimedOut,
   DirectoryUnavailable,
   ProjectNotFound,
-  ThreadNotFound,
+  TerminalLaunchFailed,
   ZmxUnavailable,
 } from "../api/contract.ts"
 import { Directories } from "../platform/directories.ts"
 import { ProjectRepository } from "../projects/projectRepository.ts"
 import { Terminals } from "../terminals/terminals.ts"
+import type { Terminal } from "../terminals/terminals.ts"
 import { ThreadRepository } from "./threadRepository.ts"
 import type { ThreadRecord } from "./threadRepository.ts"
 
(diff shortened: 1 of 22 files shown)


app-server/src/threads/threads.ts
```diff
@@ -61,129 +104,474 @@ export class Threads extends Context.Service<
...
+                Effect.succeed(Stream.empty as Stream.Stream<AgentActivity>),
+              ),
+              Scope.provide(child),
+            )
+          yield* stream.pipe(
+            Stream.runForEach((activity) =>
+              Effect.gen(function* () {
+                const previous = liveActivity.get(record.id)
+                liveActivity.set(record.id, activity)
+                if (previous !== undefined && isBusy(previous) && activity === "idle") {
```
jeremytondo: [P1] Preserve first-turn completion across observation gaps. Confirmation is written only if this process previously observed a busy state. If ATC restarts (or the provider observer reconnects) after the first turn became busy but before its idle event, the next idle arrives with `previous === undefined`, so the session remains unconfirmed even though it now has history. When the TUI later ends, `openTerminal` treats it as provisional and re-materializes a new identity, abandoning that conversation. Please carry/reconcile enough completion evidence to confirm across this gap, and add a restart-mid-first-turn regression (not only the current restart-before-any-activity case).

app-server/src/threads/threads.ts
```diff
@@ -61,129 +104,474 @@ export class Threads extends Context.Service<
...
+            }
+            const prepared = yield* adapter
+              .prepareTuiSession({ cwd: record.workingDirectory })
+              .pipe(Effect.mapError(mapAgentError(record)))
+            const terminal = yield* createTuiTerminal(record, prepared.launchSpec)
+            const cleanupTerminal = terminals
+              .delete(terminal.id)
+              .pipe(Effect.catch(() => Effect.void))
+            yield* Effect.gen(function* () {
+              const identity = yield* prepared.identity.pipe(Effect.mapError(mapAgentError(record)))
```
jeremytondo: [P2] Release a claimed identity if adoption does not complete. Awaiting Claude's `prepared.identity` flips its adapter allocation to `claimed`, so the prepared scope finalizer intentionally stops revoking the hook secret/removing the hook files. A client interruption, timeout, or defect after this line but before `setProviderSession` succeeds currently deletes only the terminal; the adapter resources survive with no thread row owning their metadata (the explicit deleted-row branch is the only path that releases them). Please bracket ownership so `releaseSession(identity)` runs on every non-success until the provider identity has actually been persisted, and add an interruption test after identity resolution but before adoption.

jeremytondo (commented):
Two actionable lifecycle gaps found in confirmation and materialization cleanup.

app-server/src/threads/threads.ts
jeremytondo: Fixed in 1227bd2, with a deliberately different rule than suggested: the confirmed marker is now persisted at the FIRST busy signal observed on the feed, not on busy→idle. A submitted prompt is already durable provider history worth protecting, and persisting at onset means there is nothing left to reconcile across a restart or observer gap. Carrying separate completion evidence would still fail when the idle lands during ATC downtime (providers broadcast status *changes*, so nothing arrives after re-subscribing) and could never recover for Claude, whose checkSession is honestly `unknown`. Added the restart-mid-first-turn regression (busy observed → restart → idle never observed by any process → the next open resumes the exact id, never re-materializes) and updated the superseded-feed test to the onset semantics.

app-server/src/threads/threads.ts
jeremytondo: Fixed in 1227bd2. Identity adoption is now bracketed inside `Effect.uninterruptibleMask`: the identity await stays interruptible (openTerminal's bound can still cancel it), but from resolution until `setProviderSession` succeeds, every non-success — deleted row, persistence defect, or interruption — releases the fresh session via an uninterruptible `onExit` (the explicit deleted-row release is folded into the same bracket). The mask leaves no yield point where interruption can observe a resolved-but-unowned identity, so a literal interrupt-between-resolution-and-adoption repro isn't constructible without a test-only seam in the adoption path; the bracket's release-on-non-success is regression-tested deterministically instead: the fake's identity gate resolves after a mid-open delete, and the fresh session must be released because no thread row will ever own it.

PR #138, opened 2026-08-06
ATC-128: client event streaming for live resource updates

jeremytondo:
# ATC-128: Client event streaming for live resource updates

Implements [ATC-128](https://linear.app/elevenideas/issue/ATC-128/client-event-streaming-for-live-resource-updates): a contract-defined SSE stream at `GET /api/v1/events` carrying thin `ResourceChangedEvent { resource, id, change }` invalidations for Projects, Threads, and Terminals, so clients stay current without polling. Payloads are Schema-typed and derive from the contract like every other surface.

## What's here

- **Contract** — `ResourceChangedEvent` schema + `subscribeEvents` endpoint (`HttpApiSchema.StreamSse`, data mode). Thin invalidation only: clients coalesce and refetch through existing GETs, which stay the single source of truth.
- **Events service** (`src/events/events.ts`) — in-process fan-out. Each subscriber gets its own bounded queue; `publish` never blocks or fails; overflow ends only the lagging subscriber's stream (its reconnect resync heals it); `close()` ends every stream cleanly.
- **Emission points** — service-layer publishes next to each successful mutation in Projects (create/update/delete), Terminals (create incl. startup-recovery markLive, rename, delete, every markEnded path: reconcile tombstoning and `confirmEnded`), and Threads (create, rename, archive/unarchive transitions, delete, openTerminal, observed activity-state changes). Repositories stay SQL-only.
- **Subscribe handler** — runs one terminals reconcile pass per new subscriber, then streams. Implemented with `handleRaw` (see decisions below).
- **Shutdown** — a `drainEventsBeforeStop` layer sequenced above `HttpRouter.serve` ends all subscriber streams before Bun's 2s bounded graceful stop, so shutdown with open subscribers completes quickly and quietly instead of riding the forced-interrupt path.
- **OpenAPI/Swift** — `openapi.ts` additionally emits the plain `ResourceChangedEvent` object schema as an ordinary component; the pinned Swift generator now produces a real `ResourceChangedEvent` struct for decoding SSE `data` (data-mode `StreamSse` alone yields only a `String` typealias).

## Key decisions made during implementation

1. **Per-subscriber bounded `Queue`s instead of a raw Effect `PubSub`.** v4 `PubSub.shutdown` *interrupts* takers (the noisy path the issue requires avoiding), and none of its overflow strategies close just the lagging subscriber. A `Set` of bounded queues (`offerUnsafe` false ⇒ `Queue.end` that subscriber) implements the issue's exact semantics with clean `Done` completion via `Stream.fromQueue`.
2. **`handleRaw` + an opening `": connected"` SSE comment.** Discovered during testing: Bun's `fetch` does not resolve a chunked response until its first body byte arrives, and this stream emits nothing until the first change — so every fetch-based client (the contract TS client included) would block in subscribe until an unrelated mutation happened, breaking subscribe-then-resync. A spec-standard SSE comment as the first bytes fixes it; conforming parsers (Effect's SSE decoder, Swift's) ignore comment lines. The typed `.handle` pipeline has no seam for it, so the handler is `handleRaw` with events encoded through the contract schema — wire shape identical to the typed pipeline (the typed test client decodes it in the tests).
3. **Drain layer above the listener, not just a service finalizer.** Layer finalizers run dependents-first; Events (a dependency of the handlers) would release *after* the serve layer's bounded graceful stop already waited out its 2s timeout on the open SSE response. The drain layer is composed as a dependent of the serve stack, so its finalizer runs first.
4. **Companion thread events for terminal transitions.** A terminal transition with a `threadId` changes the thread's derived contract fields (`linkedTerminalId`, activity re-derivation), and thread deletion nulls surviving tombstones' `threadId` via the FK — the event names no other resource, so those cross-resource changes each publish their own event (found in adversarial review).
5. **Subscribers register eagerly, before the first response byte.** `Events.subscribe` is an effect that registers the queue before returning its stream, and the handler subscribes before emitting the comment — otherwise a client that resyncs on first byte could miss events in the registration gap (found in adversarial review; it also makes the pre-stream terminals reconcile land *in* the new subscriber's feed).
6. **Publish-on-change-only for activity, including the read path.** The activity consumer publishes `thread updated` only when the observed activity changes, and `resolveActivity`'s re-derivation publishes on change too (loop-safe: the busy snapshot is gone afterward, so event-triggered rereads short-circuit).

## Verification

- New: Events service unit tests (fan-out order, overflow-drops-only-the-laggard, clean close, subscribe-after-close), typed end-to-end API test (subscribe → mutate → receive → clean end), OpenAPI doc/runtime case for `/events`, and a live shutdown test (open subscriber, event over the real wire, `Scope.close` under 1.5s, stream ends without error).
- `mise run -C app-server check` and `mise run contract:check` (incl. Swift client build) pass.

## Review

Two adversarial sub-agent reviews (correctness, simplicity) ran before this PR; their accepted findings are folded in (companion thread events, eager registration, read-path activity publish, empty-patch and repeat-confirm publish guards, fail-fast openapi.ts guards, disconnect/first-byte test coverage).

### DIFF
--- a/app-server/src/api/handlers.ts
+++ b/app-server/src/api/handlers.ts
@@ -1,13 +1,18 @@
-import { Effect } from "effect"
+import { Effect, Schema, Stream } from "effect"
+import { HttpServerResponse } from "effect/unstable/http"
 import { HttpApiBuilder } from "effect/unstable/httpapi"
-import { Api } from "./contract.ts"
+import { Api, ResourceChangedEvent } from "./contract.ts"
 import { AgentRegistry } from "../agents/agentRegistry.ts"
+import { Events } from "../events/events.ts"
 import { BuildInfo } from "../platform/buildInfo.ts"
 import { Directories } from "../platform/directories.ts"
 import { Projects } from "../projects/projects.ts"
 import { attachTerminal } from "../terminals/terminalAttach.ts"
 import { Terminals } from "../terminals/terminals.ts"
 import { Threads } from "../threads/threads.ts"
+
+/** One SSE frame per contract event: the exact wire shape StreamSse encodes. */
+const encodeEventJson = Schema.encodeEffect(Schema.fromJsonString(ResourceChangedEvent))
 
 /**
  * Implements the /api/v1 contract. App construction only — no listener, and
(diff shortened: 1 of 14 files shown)


app-server/src/terminals/terminals.ts
```diff
@@ -274,6 +285,7 @@ export const layer = Layer.effect(Terminals)(
             Effect.andThen(repository.markLive(record.id)),
             Effect.onExit((exit) => (exit._tag === "Failure" ? cleanup : Effect.void)),
           )
+        yield* events.publish({ resource: "terminal", id: live.id, change: "created" })
```
jeremytondo, line 288: **P1 — Publish the companion thread invalidation when this terminal becomes live.** When `Threads.materialize` calls `terminals.create` with a `threadId`, this live row immediately changes the thread's derived `linkedTerminalId`, but the `thread updated` event in `openTerminal` is not published until the entire identity wait finishes (up to 30 seconds). A client that refetches only the resource named by this `terminal created` event can therefore keep stale thread state. Please publish `thread updated` here whenever `input.threadId` is present; the later event can remain to cover provider-identity changes.

app-server/src/api/handlers.ts
```diff
@@ -23,46 +28,81 @@ export const V1Handlers = HttpApiBuilder.group(
...
+        // their resync. Conforming SSE parsers ignore comment lines, and
+        // events are encoded through the contract schema, so the wire shape
+        // is exactly what the typed StreamSse pipeline produces.
+        .handleRaw("subscribeEvents", () =>
+          Effect.gen(function* () {
+            // Registration first: a client may resync the moment it sees
+            // the comment, so the subscription must already be live. The
+            // reconcile pass then lands its tombstone transitions IN this
+            // subscriber's stream — the "a client showed up" refresh.
+            const feed = yield* events.subscribe()
```
jeremytondo: **P2 — Avoid registering a queue before the fallible reconciliation step.** If `terminals.list()` defects or the request is interrupted while it is running, this handler never starts `feed`, so its `Stream.ensuring` cleanup never executes. The eagerly registered queue remains in the subscriber set until 129 later events overflow it (indefinitely on an idle server), and repeated connect/abort attempts can accumulate queues. Running reconciliation before `events.subscribe()` preserves the required guarantee—the subscriber is still registered before the opening comment, and the client's post-open resync observes the reconciled state—or the registration needs an explicit request-scoped release.

app-server/test/api/api.test.ts
```diff
@@ -191,6 +195,48 @@ describe("/api/v1/projects", () => {
...
+          .subscribeEvents()
+          .pipe(
+            Effect.flatMap(Stream.runForEach((event) => Effect.sync(() => received.push(event)))),
+          ),
+      )
+      // The subscriber registers when its stream starts running; events
+      // published before that are (by design) missed, so wait for it.
+      for (let attempt = 0; (yield* events.subscriberCount()) === 0; attempt++) {
+        assert.isBelow(attempt, 100, "the subscriber never registered")
+        yield* Effect.sleep("10 millis")
```
jeremytondo: **P2 — Replace the fixed wall-clock polling with condition-driven synchronization.** These 100 × 10 ms loops can fail solely because an overloaded CI worker takes more than one second to schedule the subscriber or delivery, and they conflict with the repository invariant that tests wait on real conditions rather than sleeps. Please use a `Deferred`/condition signal or the scheduler-yielding test helper pattern. The disconnect poll in `app-server/test/server.test.ts:140-142` has the same issue.

jeremytondo (commented):
Three actionable findings from the full review.

app-server/src/terminals/terminals.ts
jeremytondo: Fixed in 633011e: `create` now publishes `thread updated` alongside `terminal created` whenever `input.threadId` is present, mirroring the delete path. The `openTerminal` event remains for provider-identity changes. The startup-recovery `markLive` path was left as-is: it runs during layer build, before the listener is up, so no subscriber can exist to receive a companion event there.

app-server/src/api/handlers.ts
jeremytondo: Fixed in 633011e by reordering: the reconcile pass now runs before `events.subscribe()`, so a defect or interrupt during `terminals.list()` leaves nothing registered. Registration still precedes the opening comment (the handler has no further yield points between subscribe and returning the response), so a client that resyncs on first byte observes the reconciled state and misses nothing — the tombstone events just land in existing subscribers' feeds instead of the new one's, which the resync covers. The residual window (interrupt landing exactly between registration and the response starting) is between adjacent effect ops with no fallible step and self-heals via the documented overflow-drop, so no request-scoped release was added.

app-server/test/api/api.test.ts
jeremytondo: Fixed in 633011e. The API test no longer polls at all: it yields `subscribeEvents()` directly — the handler registers the subscriber before producing the response, so registration is deterministic by the time the effect resolves (asserted with a plain `subscriberCount === 1`) — and delivery is awaited with a blocking `Queue.take`, bounded only by the test timeout. The `server.test.ts` disconnect poll keeps polling (unregistration rides Bun's cancel → fiber interrupt → `Stream.ensuring`, so there is no signal to await) but drops the fixed 1s cap; the test timeout is now the only bound, so a slow CI worker can no longer fail it.

PR #139, opened 2026-08-06
ATC-142: final app-server hardening before the macOS migration

jeremytondo:
Closes the gaps from the 2026-08-06 full-server readiness review so ATC-125 builds its Swift client once against a finished surface. After this lands, the contract, event stream, and attach protocol are considered stable for the macOS migration.

## Contract
- **Required `message` on every tagged-error wire encoding.** Each error class computes its message in its own constructor (the one place the rule lives); call sites are unchanged, decode recomputes it so it can never disagree with the structured fields, and Swift/CLI consumers get a printable diagnostic without duplicating message-building rules.
- **Timestamps are `format: date-time`.** Generated Swift models now carry `Date`. ATCKit gains a hand-written `ATCDateTranscoder` (lenient decode, byte-identical fractional-second encode) with an exhaustive 1000-millisecond round-trip regression test — the naive format-style encode truncated ~half of all values.
- **`subscribeEvents` description pins the race-free resync ordering** (open stream → wait for `: connected` → refetch), coalescing guidance (~100 ms, per collection), and the heartbeat cadence.

## Events stream
- **Shared ~25 s SSE heartbeat**: one timer in the Events layer ticks through the same per-subscriber queues; the handler emits it as an SSE comment. Quiet streams survive URLSession's 60 s idle teardown, clients get a liveness signal, and dead sockets are torn down by their failing write. Covered at the service level and over a real wire.
- **Cascade-publish regression tests** for the silent-sidebar-staleness class: thread delete → orphaned-terminal update, reconcile tombstone → terminal + thread updates, activity transitions (including no re-publish on repeated identical activity).

## Attach protocol
- **`packages/attach-protocol/`**: a client-consumable spec (framing, control vocabulary and ping/pong timing, close codes including 1006-is-retryable, pre-upgrade statuses, URL shape and the [1,1000] clamp, the zmx leader-resize caveat) plus machine-readable fixtures consumed by a new TS test today and by ATCKit later.

## Threads/agents
- **`ensureObserved` double-subscription race closed** (synchronous check + `Scope.forkUnsafe` + claim), plus the interruption-window claim leak, plus a guard so reads during an in-flight open cannot resubscribe to a superseded provider session.
- **openTerminal death-watch**: the identity wait races a backing-off reconciled liveness check, so a TUI that dies at launch fails in ~a second with a diagnostic naming the real cause instead of a 30 s generic timeout.
- **Codex reopen probes session existence** in `tuiLaunch` via the shared `thread/list` walk; a complete walk that omits the id fails `AgentResumeFailed` → 409 with a presentable reason, while an inconclusive walk (page cap / non-advancing cursor) launches blind rather than fabricating absence. Claude cannot be probed; the contract documents that clients must expect "open succeeded, terminal died".

## Diagnostics
- **"shutting down" log line** on SIGINT/SIGTERM, registered to run first among the shutdown finalizers, with a blackbox assertion.

## Verification
- `mise run -C app-server check` (251 tests), `mise run contract:check`, `swift test` in ATCKit (60 tests), `mise run macos:test` — all green.
- Two adversarial review passes (correctness + simplicity) were run; all majors and most minors were applied, including the Swift millisecond-truncation bug and a `detach` retryability contradiction in the close-code fixtures.

### DIFF
--- a/app-server/src/agents/codexAdapter.ts
+++ b/app-server/src/agents/codexAdapter.ts
@@ -544,21 +544,22 @@ export const layer = Layer.effect(CodexAdapter)(
       })
 
     /**
-     * Walk the paginated thread/list for one thread: the found entry,
-     * `"absent"` when a complete walk (nextCursor exhausted) omits the id,
-     * or `"unknown"` when the walk could not be completed — a repeated
-     * cursor or the page cap, either a provider that cannot make
-     * pagination progress or a pathologically long list. Callers must not
-     * treat `"unknown"` as absence.
+     * Walk one paginated thread/list population (thread/list returns only
+     * non-archived threads unless `archived: true` — the populations are
+     * disjoint): the found entry, `"absent"` when a complete walk
+     * (nextCursor exhausted) omits the id, or `"unknown"` when the walk
+     * could not be completed — a repeated cursor or the page cap, either a
+     * provider that cannot make pagination progress or a pathologically
+     * long list.
      */
-    const findThread = (state: ClientState, threadId: string) =>
+    const walkThreadList = (state: ClientState, threadId: string, archived: boolean) =>
       Effect.gen(function* () {
         let cursor: string | undefined
         for (let page = 0; page < 100; page++) {
           const reply = yield* request(
             state,
             "thread/list",
-            cursor === undefined ? {} : { cursor },
+            cursor === undefined ? { archived } : { archived, cursor },
           ).pipe(
             Effect.mapError((error) =>
               unavailable(error._tag === "RpcError" ? error.text : error.reason),
@@ -577,6 +578,24 @@ export const layer = Layer.effect(CodexAdapter)(
           cursor = next
         }
         return "unknown" as const
+      })
+
+    /**
+     * Find one thread across BOTH thread/list populations — a session
+     * archived through another Codex surface still exists and must not be
+     * reported as deleted. `"absent"` only when complete walks of both the
+     * non-archived and archived lists omit the id; `"unknown"` when either
+     * walk was inconclusive. Callers must not treat `"unknown"` as absence.
+     */
+    const findThread = (state: ClientState, threadId: string) =>
+      Effect.gen(function* () {
+        const live = yield* walkThreadList(state, threadId, false)
+        if (typeof live !== "string") return live
+        const archived = yield* walkThreadList(state, threadId, true)
+        if (typeof archived !== "string") return archived
+        return live === "absent" && archived === "absent"
+          ? ("absent" as const)
+          : ("unknown" as const)
       })
 
     const adapter: AgentAdapter = {
--- a/app-server/src/agents/codexAdapter.ts
+++ b/app-server/src/agents/codexAdapter.ts
@@ -544,16 +544,17 @@ export const layer = Layer.effect(CodexAdapter)(
       })
 
     /**
-     * Walk the paginated thread/list for one thread. `undefined` after a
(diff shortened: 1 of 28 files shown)


app-server/src/agents/codexAdapter.ts
```diff
@@ -543,6 +543,42 @@ export const layer = Layer.effect(CodexAdapter)(
...
+     * treat `"unknown"` as absence.
+     */
+    const findThread = (state: ClientState, threadId: string) =>
+      Effect.gen(function* () {
+        let cursor: string | undefined
+        for (let page = 0; page < 100; page++) {
+          const reply = yield* request(
+            state,
+            "thread/list",
+            cursor === undefined ? {} : { cursor },
```
jeremytondo: [P1] Include archived Codex threads before concluding that the session is absent

`thread/list` defaults `archived` to false/null, which the pinned Codex schema defines as “only non-archived threads.” A session archived through another Codex surface is therefore omitted from this otherwise complete walk, causing `tuiLaunch` to report it as deleted/pruned and permanently refuse the reopen even though the provider still has it. Please use the exact `thread/read` probe, or also complete an `archived: true` walk, and only classify the id as absent after both populations have ruled it out.

packages/attach-protocol/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+## URL
+
+```
+ws://<host>/api/v1/terminals/{terminalId}/attach?cols=<n>&rows=<n>
+```
+
+- Scheme: `ws:` for `http:` base URLs, `wss:` for `https:`.
+- `cols`/`rows` set the initial PTY size. Both are optional; a missing or
+  malformed value falls back to the server default (80×24). Values are
```
jeremytondo: [P2] Make malformed query values actually take the documented fallback

The current `clampDimension` uses `Number.parseInt`, so partially numeric strings are accepted: for example `cols=80junk` becomes 80 and `cols=1e2` becomes 1 instead of falling back. The new fixture only exercises a wholly nonnumeric value (`"wide"`), so it does not catch this mismatch. Please validate the entire query string before conversion (and add a numeric-prefix fixture), or document the prefix-parsing behavior instead.

jeremytondo (commented):
Two correctness/protocol mismatches to address before merging.

app-server/src/agents/codexAdapter.ts
jeremytondo: Fixed in 51f6c5e. `findThread` now walks both thread/list populations — non-archived first, then `archived: true` (the pinned schema confirms they are disjoint) — and reports `absent` only when complete walks of both omit the id; an inconclusive walk in either population stays `unknown`. Kept the list walk rather than switching to `thread/read`: a completed walk is positive proof of absence, whereas distinguishing a not-found `thread/read` error from a transient one would rely on untyped RPC error text, which is exactly the guesswork the probe was designed to avoid. The fake listener now honors the `archived` filter and implements `thread/archive`, and the test archives a thread externally and asserts checkSession still sees it and tuiLaunch still hands back the resume argv.

packages/attach-protocol/README.md
jeremytondo: Fixed in 51f6c5e. `clampDimension` now validates the entire string: `Number(raw)` with an explicit empty/whitespace guard, so `80junk` and `12junk` take the fallback instead of prefix-parsing, while whole-string numeric forms (`1e2` → 100) still clamp. Added fixtures for the numeric-prefix, scientific-notation, and whitespace-only cases, and documented the whole-string rule in the README and the fixture description.

PR #140, opened 2026-08-07
ATC-143: add server-side thread pins

jeremytondo:
## Summary

- persist an optional `pinnedAt` timestamp on threads through an append-only migration
- add idempotent `pinThread` and `unpinThread` HTTP actions
- clear pin state atomically when archiving and reject attempts to pin archived threads
- publish thread `updated` events for pin changes on the existing SSE stream
- regenerate OpenAPI and cover the generated Swift client surface

## Why

Thread pin state needs to be server-owned so every client shares the same ordering and archived threads cannot silently regain stale pin state.

## Validation

- `TMPDIR=/tmp mise run contract:check`
  - App Server: 251 passed, 10 skipped
  - ATCKit: 61 passed

[Linear: ATC-143](https://linear.app/elevenideas/issue/ATC-143/server-side-thread-pins)


app-server/src/threads/threadRepository.ts
```diff
@@ -195,7 +200,24 @@ export const layer = Layer.effect(ThreadRepository)(
...
+
+    const setPinnedRows = SqlSchema.findAll({
+      Request: Schema.Struct({
+        id: Schema.String,
+        pinned_at: Schema.NullOr(Schema.String),
+        updated_at: Schema.String,
+      }),
+      Result: ThreadRow,
+      execute: (patch) => sql`
+        UPDATE threads SET pinned_at = ${patch.pinned_at}, updated_at = ${patch.updated_at}
```
jeremytondo: **Minor (race):** `pin()` checks `archivedAt` on a record read in `threads.ts`, then this UPDATE writes `pinned_at` unconditionally. An `archiveThread` that lands between the read and this write produces an archived **and** pinned row — exactly the state the new `threads.ts` header says no client can observe. The archive side already guards this atomically in SQL (the `CASE` in `setArchivedRows`); mirroring it here closes the gap, and under the race `pin` then correctly returns the archived, unpinned thread.

```suggestion
        UPDATE threads SET
          pinned_at = CASE WHEN archived_at IS NOT NULL THEN NULL ELSE ${patch.pinned_at} END,
          updated_at = ${patch.updated_at}
```

Low urgency given the single-user loopback setup — flagging mainly because the header states the invariant unconditionally.

PR #147, opened 2026-08-09
Remote access: bind setting + bearer-token trust (ATC-148)

jeremytondo:
Implements [ATC-148](https://linear.app/elevenideas/issue/ATC-148/remote-access): the server half of secure remote access. A `bind` setting opens the listener beyond loopback, and a server-generated bearer token gates every non-loopback request. Intended posture is tailnet-only reachability (Tailscale); the token is the just-in-case backstop.

## What's here

- **`bind` config setting** — precedence flags (`--bind`) > `ATC_BIND` > `config.toml` > default `127.0.0.1`. Replaces the hardcoded loopback hostname in `server.ts`. Any address is accepted, not validated; `0.0.0.0` is the documented non-loopback value (a single non-loopback address drops the loopback listener and breaks local clients). Startup/`status` lines show the resolved address, and the pidfile now records the bind.
- **Bearer token** (`platform/authToken.ts`) — one static token, generated on first server start into a `0600` file in the data dir (SSH-host-key style). Atomic temp+rename write guarantees the mode and never exposes a partial token; `ensure` uses exclusive-create so a first-start race (`atc token` vs boot) can't mint two tokens. `verify` re-reads the file per check, so rotation is immediate with no cache to invalidate.
- **One trust rule** (`api/localTrust.ts`, global middleware over API + SSE + attach WebSocket + hook webhook): a request passes if it's a **loopback TCP peer** whose `Host`/`Origin` are loopback, **or** presents `Authorization: Bearer <token>` (timing-safe); else the same empty 403 as before. Gating the token-free path on the peer address — not the client-controlled `Host` header — is what stops a remote client sending `Host: localhost` to skip the token. Requests proxied by `tailscale serve` keep the incoming (non-loopback) `Host`, making them indistinguishable from DNS rebinding, so they present the token like a direct tailnet/LAN peer — only direct loopback traffic is token-free.
- **`atc token` / `atc token rotate`** — local filesystem commands over the data dir (no server needed, no connection flags), print and reissue the credential.
- **No contract/OpenAPI change** — auth is a property of how you reach the server, not of any endpoint.
- **macOS** — fixed the stale `127.0.0.1:7331` placeholder in `ConnectionsSettingsView` (default port is 7332). No new client machinery (ATC-125 already shipped the Keychain token + `BearerAuthMiddleware` seams).

## Key decisions

- **Peer-address gating over Host-only trust.** The original loopback guard keyed trust off the `Host` header, which was safe only because the listener bound `127.0.0.1`. Opening `bind` made that a token-bypass (`Host: localhost` from any remote peer). The fix gates the token-free path on the real TCP peer (`request.remoteAddress`). `tailscale serve` still works — its proxied requests carry the bearer header remote clients attach anyway (the proxy preserves the incoming `Host`, so token-free proxying would need a Host allowlist nothing requires). Verified live: a remote peer with `Host: localhost` and no token now 403s.
- **Broken token file never blocks boot.** A loopback-only user who never wants remote access shouldn't be stopped by a bad `auth-token`; the layer logs a warning and continues, and `verify` fails closed.
- **Rotation is lazy re-read, not restart-required** — the simplest reliable rule; the per-request file read is paid only by non-loopback tokened requests.

## Testing

- Unit: token generate/ensure/rotate/verify (0600, rotation-follows-file, fail-closed), `isLoopbackAddress` spoof-defense table, `bind` config precedence.
- In-process listener: non-loopback `Host` + valid token passes, missing/wrong/mis-schemed token 403s, loopback free, attach WebSocket path under the same guard.
- Blackbox: token file creation + `0600`, `atc token`/`rotate` stdout, live rotation enforcement against a real `atc serve`, `serve --bind 0.0.0.0` serving over loopback.
- `mise run -C app-server check` green (266 passed). Manual macOS end-to-end over the tailnet remains the acceptance step per the issue.

## Review

Two adversarial sub-agent reviews (correctness + simplicity). The correctness pass caught the Host-spoof token bypass (now fixed with peer-gating) plus non-atomic token writes, the first-start race, fail-closed boot, and the `atc start` probe hardcoding loopback — all addressed. Simplicity pass drove the pidfile/message fixes, README de-duplication, and the `--bind` blackbox test.

## Review follow-up

Commit 0b25ce6a addresses all six findings from the inline review: token-file format/permission enforcement (fail-closed verify, actionable `ensure` diagnostics, 0600 repair), no unpersisted tokens on create races, pre-ATC-148 pidfile compatibility, bearer-authenticated management probes (verified live with `--bind 192.168.1.26`), bracketed IPv6 URLs, and corrected `tailscale serve` docs per the above.

### DIFF
--- a/app-server/src/api/localTrust.ts
+++ b/app-server/src/api/localTrust.ts
@@ -28,9 +28,15 @@ import * as AuthToken from "../platform/authToken.ts"
 // The token-free path is gated on the TCP PEER address, never on the `Host`
 // header alone: `Host` is client-controlled, so trusting a "loopback" Host
 // from a non-loopback peer would let any remote client send `Host: localhost`
-// and skip the token entirely. Peer-gating also keeps `tailscale serve`
-// working token-free — it proxies to us over loopback — while a direct
-// tailnet/LAN connection (a non-loopback peer) must present the token.
+// and skip the token entirely. The loopback-Host requirement still applies
+// to loopback peers because a DNS-rebinding request IS a loopback peer
+// carrying the hostile name in `Host` — and a local reverse proxy
+// (`tailscale serve` preserves the incoming Host) presents exactly that same
+// shape, indistinguishable from rebinding here. Proxied requests therefore
+// authenticate with the bearer token like any other remote path; making them
+// token-free would take a configured Host allowlist, which nothing needs
+// while every remote client attaches the token anyway (remote browsers are
+// out of scope by design).
 
 const LOOPBACK_NAMES = new Set(["localhost", "127.0.0.1", "[::1]"])
 
--- a/app-server/src/api/localTrust.ts
+++ b/app-server/src/api/localTrust.ts
@@ -1,19 +1,46 @@
 import { Effect, Option } from "effect"
 import { HttpRouter, HttpServerRequest, HttpServerResponse } from "effect/unstable/http"
+import * as AuthToken from "../platform/authToken.ts"
 
-// Local trust for the loopback listener (ATC-121): the server only ever binds
-// 127.0.0.1, and this middleware rejects the one attack class that still
-// reaches a loopback control plane from a browser — DNS rebinding (a hostile
-// name resolving to 127.0.0.1 puts its own name in Host) and CSRF (a hostile
+// The trust rule, in one sentence: a request passes if it meets the loopback
+// Host/Origin rules below OR presents the bearer token (ATC-148); everything
+// else gets the same empty 403. One global middleware, so the API, SSE, the
+// attach WebSocket upgrade, and the hook webhook are covered uniformly.
+//
+// Loopback trust (ATC-121): the default listener binds 127.0.0.1, and the
+// Host/Origin rules reject the one attack class that still reaches a
+// loopback control plane from a browser — DNS rebinding (a hostile name
+// resolving to 127.0.0.1 puts its own name in Host) and CSRF (a hostile
 // page's request carries its Origin). Requests must present a recognized
 // loopback Host, and browser requests (Origin present) must come from a
 // loopback origin. Non-browser clients send no Origin and pass untouched.
 // Origin-less browser requests (e.g. a cross-site <img>) can only be simple
 // GETs — safe while every GET is side-effect-free and responses are
 // unreadable cross-origin; revisit if a GET ever gains a side effect.
-// The remote-auth pass later relaxes this deliberately for tokened clients.
+//
+// Token trust (ATC-148): with the `bind` setting opened beyond loopback
+// (intended posture: tailnet-only reachability), non-loopback requests pass
+// only with `Authorization: Bearer <token>` matching the persisted token
+// (platform/authToken.ts, timing-safe). Loopback stays token-free: local
+// clients ("no secrets, just the URL") are untouched. Browsers cannot attach
+// bearer headers to SSE/WebSockets, so remote browser access 403s by design.
+//
+// The token-free path is gated on the TCP PEER address, never on the `Host`
+// header alone: `Host` is client-controlled, so trusting a "loopback" Host
+// from a non-loopback peer would let any remote client send `Host: localhost`
+// and skip the token entirely. Peer-gating also keeps `tailscale serve`
+// working token-free — it proxies to us over loopback — while a direct
+// tailnet/LAN connection (a non-loopback peer) must present the token.
 
 const LOOPBACK_NAMES = new Set(["localhost", "127.0.0.1", "[::1]"])
+
+// Peer addresses Bun reports for a loopback connection: IPv4, IPv6, and the
+// IPv4-mapped-IPv6 form. Any 127.0.0.0/8 address is loopback too.
+export const isLoopbackAddress = (address: string): boolean =>
+  address === "::1" ||
+  address === "::ffff:127.0.0.1" ||
+  address.startsWith("127.") ||
+  address.startsWith("::ffff:127.")
 
 /** `Host` header values that identify this loopback server (optional port). */
 export const isLoopbackHost = (host: string | undefined): boolean => {
@@ -37,9 +64,10 @@ export const isLoopbackOrigin = (origin: string): boolean => {
 }
 
 /**
- * Global route middleware: reject spoofed-`Host`/cross-`Origin` requests with
- * an empty 403, and annotate every log line in an accepted request with the
- * request span ids so file logs correlate per request.
+ * Global route middleware: reject requests that neither satisfy the loopback
+ * `Host`/`Origin` rules nor present the bearer token with an empty 403, and
+ * annotate every log line in an accepted request with the request span ids
+ * so file logs correlate per request.
  */
 export const middleware = HttpRouter.middleware(
   Effect.fnUntraced(function* <E, R>(
@@ -49,12 +77,16 @@ export const middleware = HttpRouter.middleware(
       HttpServerRequest.HttpServerRequest | R
     >,
   ) {
+    const authToken = yield* AuthToken.AuthToken
     const request = yield* HttpServerRequest.HttpServerRequest
     const origin = request.headers["origin"]
-    if (
-      !isLoopbackHost(request.headers["host"]) ||
-      (origin !== undefined && !isLoopbackOrigin(origin))
-    ) {
+    // A loopback PEER (unknown peer fails closed) whose Host/Origin also pass
+    // is the token-free local path; every other request needs the token.
+    const loopback =
+      Option.match(request.remoteAddress, { onNone: () => false, onSome: isLoopbackAddress }) &&
+      isLoopbackHost(request.headers["host"]) &&
+      (origin === undefined || isLoopbackOrigin(origin))
+    if (!loopback && !(yield* authToken.verify(request.headers["authorization"]))) {
       return HttpServerResponse.empty({ status: 403 })
     }
     const span = yield* Effect.option(Effect.currentParentSpan)
--- a/app-server/src/cli/serve.ts
+++ b/app-server/src/cli/serve.ts
@@ -27,8 +27,13 @@ export const Port = Schema.Int.check(Schema.isBetween({ minimum: 1, maximum: 655
 export const port = Flag.integer("port").pipe(
   Flag.withSchema(Port),
   Flag.optional,
+  Flag.withDescription("TCP port to listen on (overrides ATC_PORT/config.toml/default)"),
+)
+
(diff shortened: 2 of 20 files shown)


jeremytondo (commented):
Full review: I found five substantive correctness/security issues and one IPv6 diagnostics issue. The full app-server gate passes (266 tests), but the items below need attention before merge.

app-server/src/api/localTrust.ts
```diff
@@ -49,12 +77,16 @@ export const middleware = HttpRouter.middleware(
...
     const origin = request.headers["origin"]
-    if (
-      !isLoopbackHost(request.headers["host"]) ||
-      (origin !== undefined && !isLoopbackOrigin(origin))
-    ) {
+    // A loopback PEER (unknown peer fails closed) whose Host/Origin also pass
+    // is the token-free local path; every other request needs the token.
+    const loopback =
+      Option.match(request.remoteAddress, { onNone: () => false, onSome: isLoopbackAddress }) &&
+      isLoopbackHost(request.headers["host"]) &&
```
jeremytondo, line 93: **P1 — `tailscale serve` is not token-free with this condition.** Tailscale's HTTP reverse proxy preserves the incoming host for normal backends (`r.Out.Host = r.In.Host`): https://github.com/tailscale/tailscale/blob/e592a0c36398be8a435d1cc9e039c732c540051c/ipn/ipnlocal/serve.go#L953-L979. A request to `https://workstation.<tailnet>.ts.net` therefore reaches ATC from a loopback peer but retains a non-loopback `Host`, so this expression is false and the request gets 403 without a bearer token. The new black-box test's loopback connection plus remote `Host` reproduces that exact shape and currently expects 403. This contradicts the PR's stated `tailscale serve` behavior; the proxy trust rule needs to account for the preserved host while retaining DNS-rebinding protection.

app-server/src/cli/serve.ts
```diff
@@ -80,8 +89,8 @@ export const serve = Command.make("serve", { port }, ({ port }) =>
 
 // --- Background management ---
 
-/** Pidfile contents: which process, and which port it was started on. */
-const PidRecord = Schema.Struct({ pid: Schema.Int, port: Schema.Int })
+/** Pidfile contents: which process, and where it was told to listen. */
+const PidRecord = Schema.Struct({ pid: Schema.Int, port: Schema.Int, bind: Schema.String })
```
jeremytondo: **P1 — this breaks background-server upgrades.** The previous release wrote `{ pid, port }`; making `bind` required causes those existing pidfiles to fail decoding, and `readPidRecord` converts that failure to `undefined`. After upgrading, `stop`/`status` report “not running” while the old process remains alive, and `start` then refuses because the health endpoint is occupied. Please decode the old shape compatibly and default a missing `bind` to `127.0.0.1`.

app-server/src/platform/authToken.ts
```diff
@@ -0,0 +1,129 @@
...
+    yield* fs.rename(temp, file)
+    return token
+  })
+
+/** Read the persisted token, generating and persisting one if absent. */
+export const ensure = Effect.gen(function* () {
+  const config = yield* AppConfig
+  const fs = yield* FileSystem.FileSystem
+  const existing = yield* readToken(fs, config.tokenFile)
+  if (existing !== undefined) return existing
```
jeremytondo: **P1 — existing token files bypass the advertised credential guarantees.** Any nonempty text is accepted here without validating the `atc_` format/entropy or checking/repairing `0600` permissions. A restored or accidentally created `0644` file containing a weak value therefore becomes a remotely accepted, world-readable bearer credential. I verified that changing a valid token file to `0644` and running `atc token` leaves it at `0644`. Please validate existing contents and enforce the permissions before returning them (or fail/rotate with an actionable diagnostic).

app-server/src/platform/authToken.ts
```diff
@@ -0,0 +1,129 @@
...
+  // `atc token` on a fresh install) already created the file, adopt the
+  // winner's token instead of overwriting it — otherwise each caller could
+  // mint and hand out a different token.
+  return yield* fs
+    .writeFileString(config.tokenFile, `${token}\n`, { flag: "wx", mode: 0o600 })
+    .pipe(
+      Effect.as(token),
+      Effect.catchTag("PlatformError", (error) =>
+        error.reason._tag === "AlreadyExists"
+          ? readToken(fs, config.tokenFile).pipe(Effect.map((winner) => winner ?? token))
```
jeremytondo: **P2 — this can return a credential that was never persisted.** An empty existing file is mapped to `undefined`; the exclusive create then gets `AlreadyExists`; this second read is also `undefined`; and `winner ?? token` returns the caller's newly generated value while leaving the file empty. I reproduced `atc token` exiting 0 and printing an `atc_…` value while `auth-token` remained zero bytes. The same window exists when concurrent `ensure` calls race after the winner creates the inode but before it finishes writing. Please retry reading/fail rather than returning a token that is not on disk.

app-server/src/cli/serve.ts
```diff
@@ -92,10 +101,17 @@ const readPidRecord = (fs: FileSystem.FileSystem, file: string) =>
...
+// itself from the same host. IPv6 literals need brackets in a URL.
+const reachableHost = (bind: string): string =>
+  bind === "0.0.0.0" || bind === "::" || bind === "" ? "127.0.0.1" : bind
+const hostForUrl = (host: string): string => (host.includes(":") ? `[${host}]` : host)
+
+/** One bounded health probe against `host`; false on any failure. */
+const probeHealth = (host: string, port: number, timeoutMillis: number) =>
   Effect.tryPromise(() =>
-    fetch(`http://127.0.0.1:${port}/api/v1/health`, {
+    fetch(`http://${hostForUrl(host)}:${port}/api/v1/health`, {
```
jeremytondo: **P2 — `atc start --bind <specific-non-loopback-address>` cannot pass this probe.** `reachableHost` returns the specific address, so this request carries a non-loopback `Host`; the trust middleware requires the bearer token, but the probe sends none. The running server returns 403, `start` waits 15 seconds and kills it, and `status` also reports it unhealthy. Either authenticate management probes, probe through loopback when the bind includes it, or reject bind values the lifecycle commands cannot support.

app-server/src/cli/serve.ts
```diff
@@ -60,11 +69,11 @@ export const serve = Command.make("serve", { port }, ({ port }) =>
     if (build.commit !== "dev") {
       yield* Effect.sync(() => {
         process.stderr.write(
-          `atc app server starting on http://127.0.0.1:${effective.port} (logs: ${effective.logFile})\n`,
+          `atc app server starting on http://${effective.bind}:${effective.port} (logs: ${effective.logFile})\n`,
```
jeremytondo: **P3 — bracket IPv6 literals in displayed URLs.** For a bind such as `::1`, this prints `http://::1:7332`, which is not a valid URL. `hostForUrl` below already implements the required brackets; please share it here and in `server.ts`'s listen-address log.

app-server/src/api/localTrust.ts
jeremytondo: Fixed in 0b25ce6a by correcting the claim, not the condition. A proxied request (loopback peer, preserved non-loopback `Host`) is server-side indistinguishable from DNS rebinding — under rebinding the page becomes same-origin, so every header except `Host` is attacker-controlled, and `Host` is exactly what both shapes share. The only way to keep `tailscale serve` token-free would be a configured Host allowlist, which nothing needs: remote browser access is an explicit ATC-148 non-goal, and every intended remote client (macOS `BearerAuthMiddleware`, future remote CLI) attaches the bearer header — through the proxy the token passes straight through, so `tailscale serve` works, with the token, and adds HTTPS rather than an auth exemption. The module header, README, and PR body now state this; the black-box test's 403 expectation stands as the intended behavior.

app-server/src/cli/serve.ts
jeremytondo: Fixed in 0b25ce6a: `bind` is now `Schema.optionalKey` and `readPidRecord` defaults a missing bind to `127.0.0.1` (what the pre-ATC-148 server was hardcoded to listen on). Added a black-box test that writes an old-shape `{ pid, port }` pidfile and asserts `atc stop` decodes it (removes the stale file rather than reporting a failed-decode "not running").

app-server/src/platform/authToken.ts
jeremytondo: Fixed in 0b25ce6a. `ensure` now refuses any contents `generate` could not have produced (`atc_` + 43 base64url chars) with an actionable diagnostic ("delete it or run `atc token rotate`") instead of adopting them, and re-asserts 0600 on an existing valid file; `verify` applies the same format gate so a malformed stored value fails closed even though the boot layer tolerates the broken file. Verified live: a weak 0644 file makes `atc token` exit 1 with the diagnostic, and a valid file chmodded to 0644 is back at 0600 after the next `atc token`. Unit tests cover all three paths.

app-server/src/platform/authToken.ts
jeremytondo: Fixed in 0b25ce6a: the AlreadyExists branch now returns the winner's token only when it reads back as a valid token; anything else (empty file, mid-write race, corrupt contents) fails with the same actionable TokenFileError rather than handing out a credential that is not on disk. The empty-file repro now exits 1 with the diagnostic and rerunning after `atc token rotate` recovers. Covered by a unit test.

app-server/src/cli/serve.ts
jeremytondo: Fixed in 0b25ce6a by authenticating the management probes: when the probe host is not a loopback name under the trust rule, `start` prepares the token up front via `ensure` (the spawned server adopts the same file — the exclusive-create race handling covers this) and both its probes send `Authorization: Bearer`; `status` does the same with a read-only accessor so it never creates files. Loopback probes stay token-free, so a broken token file still cannot block a loopback start (mirrors the server's own boot rule). Verified live on a real interface: `atc start --bind 192.168.1.26` becomes healthy, an unauthenticated probe of the same endpoint still 403s, and `status`/`stop` work. Automated coverage uses `--bind ::1` (a specific non-wildcard bind portable across CI and macOS); a routable-address bind isn't portable in CI, hence the live verification.

app-server/src/cli/serve.ts
jeremytondo: Fixed in 0b25ce6a: `hostForUrl` moved to `server.ts` (exported) and is now used by the serve startup stderr line, the listen-address log, and all the CLI URLs. The new `--bind ::1` black-box test asserts `http://[::1]:<port>` in both `start` and `status` output.

PR #152, opened 2026-08-10
fs/list contract + server endpoint (ATC-152)

jeremytondo:
Server half of ATC-151 (server-backed folder browser). Adds `GET /api/v1/fs/list`: lists a directory's subdirectories — non-recursive, dotfolders excluded, symlinks included when they resolve to directories, sorted case-insensitively (pinned collation) — with the same bounded 2s timeout and tagged errors as `fs/check`. Omitting `path` lists the server's home directory, which is now settled on `AppConfig.home` ($HOME, else the OS account database) per the everything-reads-AppConfig invariant.

- `Directories.list` in `platform/directories.ts`; the timeout wrapper and errno→state mapping are shared between check and list. Unknown dirent types (d_type-less filesystems like NFS) are stat-classified in parallel rather than dropped.
- `atc fs list [path]` CLI command (relative paths resolve client-side, like `fs check`).
- The two macOS `APIProtocol` test doubles gain `listDirectory` conformance stubs **in this PR** so macOS CI (which triggers on `openapi.json`) stays green standalone; the picker itself is the stacked PR.
- Regenerated OpenAPI artifact; contract:check green; new API, OpenAPI-conformance, and CLI blackbox tests. The omitted-path test lists a scratch fixture home, never the runner's real $HOME.

Linear: ATC-152 (parent ATC-151). Stacked: #153 builds on this branch.

### DIFF
--- a/app-server/src/platform/directories.ts
+++ b/app-server/src/platform/directories.ts
@@ -1,11 +1,15 @@
 import { Context, Effect, Layer } from "effect"
 import { constants } from "node:fs"
 import * as fs from "node:fs/promises"
+import { dirname, join } from "node:path"
 import { DirectoryCheckTimedOut, DirectoryUnavailable, DirectoryState } from "../api/contract.ts"
+import { AppConfig } from "./config.ts"
 
-// Demand-driven directory health (ATC-121, Domain Model rules): checks run
-// only when an operation needs a directory or a client asks explicitly —
-// results are never persisted and normal reads never touch the filesystem.
+// Demand-driven directory health (ATC-121, Domain Model rules) and the
+// read-only subdirectory listing behind the folder browser (ATC-151): checks
+// and listings run only when an operation needs a directory or a client asks
+// explicitly — results are never persisted and normal reads never touch the
+// filesystem.
 // Validation requires an existing directory with read+traversal access
 // (deliberately not write: testing writability at registration is worse than
 // surfacing a write failure at first use). Identity is the symlink-resolved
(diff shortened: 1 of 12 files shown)


app-server/src/platform/directories.ts
```diff
@@ -82,37 +94,130 @@ const probe = (path: string): Promise<Probe> =>
...
+        dirent.isCharacterDevice() ||
+        dirent.isBlockDevice()
+      ) {
+        continue
+      }
+      unclassified.push({ name: dirent.name, path: entryPath })
+    }
+    // Stat the unclassified entries together: d_type-less filesystems report
+    // whole directories this way, and the timeout budget is fixed.
+    const classified = await Promise.all(
```
jeremytondo: Could we bound this `stat` fan-out? On a `d_type`-less filesystem every entry lands here, so a large directory launches one `fs.stat` promise per entry at once. `Effect.timeoutOrElse` returns after 2s but does not cancel these Node promises, meaning repeated requests can leave thousands of stats running and exhaust resources after the API has already timed out. A small worker pool or concurrency-limited traversal would preserve the bounded response while also bounding the underlying work.

jeremytondo (commented):
Quick review: one reliability issue to address in the directory-listing path.

app-server/src/platform/directories.ts
jeremytondo: Done in e9bb7d5 — the unclassified entries now go through a fixed worker pool (STAT_CONCURRENCY = 16), so a d_type-less directory holds at most 16 stats in flight and a timed-out request abandons bounded work instead of thousands of promises.

PR #153, opened 2026-08-10
macOS DirectoryPickerSheet replacing NSOpenPanel (ATC-153)

jeremytondo:
macOS half of ATC-151, stacked on #152 (needs the regenerated Swift client for `fs/list`).

Browse… now opens a server-backed folder browser driven by `GET /fs/list` through the selected Connection, so the browsed filesystem is always the server's — loopback or remote, one code path. `NSOpenPanel` is removed entirely.

- `DirectoryPickerSheet` + `DirectoryPickerModel`: drill-down list (path bar with Up, descend on click, Choose/Cancel via the shared `SheetScaffold`, which gained a non-Form content mode). One request in flight at a time; a failed load keeps the last good listing and shows the server's tagged diagnostic; a stale starting path falls back to the server home.
- `WorkingDirectoryField`: Browse… disabled with no Connection; the picker starts at the typed path when its check is available; a Connection change dismisses the picker; the existing `/fs/check` flow still validates the chosen value. Covers all three creation sheets.
- Preview + scriptable test clients implement the new operation; new `DirectoryPickerModel` tests.

Linear: ATC-153 (parent ATC-151). Merge #152 into main first, then retarget/merge this.

### DIFF
--- a/macos/ATC/Shared/DirectoryPickerSheet.swift
+++ b/macos/ATC/Shared/DirectoryPickerSheet.swift
@@ -1,0 +1,163 @@
+// Server-backed folder browser (ATC-151): Browse… always lists the server's
+// filesystem through `GET /fs/list` on the selected Connection — loopback or
+// remote, one code path. Directories only, dotfolders excluded server-side;
+// the text field stays the escape hatch for hidden paths. Read-only: no New
+// Folder, no file selection.
+
+import ATCAppServerAPI
+import SwiftUI
+
+@MainActor @Observable
+final class DirectoryPickerModel {
+    private let client: any APIProtocol
+
+    private(set) var listing: Components.Schemas.FsListResponse?
+    private(set) var errorMessage: String?
+    private(set) var isLoading = false
+
+    init(client: any APIProtocol) {
+        self.client = client
+    }
+
+    /// The canonical path Choose commits: the directory currently listed.
+    var currentPath: String? { listing?.path }
+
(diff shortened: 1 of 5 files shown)


macos/ATC/Shared/DirectoryPickerSheet.swift
```diff
@@ -0,0 +1,160 @@
...
+        self.onChoose = onChoose
+        _model = State(initialValue: DirectoryPickerModel(client: client))
+    }
+
+    var body: some View {
+        SheetScaffold(
+            title: "Choose Folder",
+            systemImage: "folder",
+            primaryLabel: "Choose",
+            canSubmit: model.currentPath != nil,
```
jeremytondo: Please disable submission while navigation is in flight (`model.currentPath != nil && !model.isLoading`). The model intentionally keeps the last good listing during a load, so after clicking a child folder the Choose button still commits the parent until the response arrives. On a slow or remote connection, a user can therefore choose a different folder than the row they just clicked.

jeremytondo (commented):
Quick review: one user-visible race in the directory picker.

macos/ATC/Shared/DirectoryPickerSheet.swift
jeremytondo: Done in 6fb574c — canSubmit now requires !model.isLoading, so Choose is disabled while a navigation is in flight and can no longer commit the folder the user just left.

PR #157, opened 2026-08-10
Auto-name new threads, with manual names always winning (ATC-155)

jeremytondo:
Implements [ATC-155](https://linear.app/elevenideas/issue/ATC-155/auto-name-new-threads-with-manual-names-always-winning): unnamed threads get a short generated title keyed off their first user prompt; a creation-time or manual name always wins and permanently suppresses auto-naming.

## How it works

- **Seam** (`agentAdapter.ts`): `observeSession` now streams `AgentSessionEvent` (activity + best-effort `userPrompt` evidence), and adapters gain a `generateTitle` one-shot. The shared `titleInstruction` template and `sanitizeTitle` output hygiene (first line, quotes stripped, ~50-char word-boundary cap) live next to the seam so both providers' titles read the same.
- **Claude**: the `UserPromptSubmit` webhook payload carries the prompt (root only — `agent_id` payloads are subagent dispatch, excluded). `generateTitle` is one SDK `query()` one-shot: `maxTurns: 1`, no tools, `persistSession: false`, `model: "haiku"` (alias verified live), 90s bound, abort-on-scope-close.
- **Codex**: probed live — `item/*` notifications never fan out to passive sockets, so the first busy transition forks a single-flight `thread/read {includeTurns: true}` against rollout history and emits the first `userMessage` text; the activity feed (and the confirmed-marker write it drives) never waits on the read. Hard read failures spend the one attempt; empty history retries boundedly. `generateTitle` spawns `codex exec --ephemeral --skip-git-repo-check -s read-only` (low reasoning) with the prompt on stdin and the reply via `--output-last-message`.
- **Threads**: the observation consumer fires one fire-and-forget generation per subscription when the thread was unnamed AND unconfirmed at subscription start. Adoption goes through the new guarded repository write — `UPDATE … SET name WHERE id = ? AND name IS NULL RETURNING *` — so a manual rename landing mid-generation wins atomically (T3Code semantics, not OpenCode's clobber). A publish fires only when a row actually changed; every failure logs debug and leaves the fallback name.
- **Subprocess**: gains `unsetEnv` so the nested-session marker scrub happens inside the seam (no `process.env` reads in the adapter).

No contract changes — `name` and the rename pipeline already existed, so clients get this for free over the existing SSE invalidate-and-refetch loop.

## Review

Reviewed by Codex (Sol, high reasoning); all four should-fix findings addressed: prompt discovery no longer blocks the activity feed, failure/empty reads are discriminated with bounded retries, a synchronously-throwing SDK setup is a typed error, and the env boundary moved into the Subprocess seam. The exec test now pins the safety-relevant invocation shape (ephemeral, read-only, stdin transport, marker scrub, temp-file cleanup).

## Testing

- `mise run -C app-server check` — fmt, typecheck, 326 tests, OpenAPI drift, all green.
- New coverage: threads policy (title lands + publishes, creation-time name suppresses, rename-mid-generation wins, failure silent, guarded-write atomicity incl. vanished rows), Claude webhook prompt capture + titler success/failure/sync-throw, Codex prompt-read exactly-once + never-blocks-activity (delayed-read fixture) + titler success/failure, sanitizer units.

https://claude.ai/code/session_01NKUGAXvzdkAat2SigtLS8W


app-server/src/agents/claudeHooks.ts
```diff
@@ -195,7 +196,22 @@ const HookPayload = Schema.Struct({
...
+ * The ROOT session's user prompt from a UserPromptSubmit payload, or null:
+ * a payload carrying `agent_id` fired inside a subagent — its prompt is
+ * synthetic dispatch, never the user's message (ATC-155).
+ */
+export const userPromptOf = (
+  eventName: string,
+  payload: Record<string, unknown>,
+): string | null => {
+  if (eventName !== "UserPromptSubmit") return null
+  if (typeof payload["agent_id"] === "string") return null
```
jeremytondo, line 211: **[P2] Ignore machine-authored root prompts.** `agent_id` excludes subagents, but Claude's `UserPromptSubmit` payload can also identify root-level `system`, `loop_wakeup`, and `schedule_wakeup` sources. If one of those is the first observed event, it consumes the one naming attempt and titles the thread from a machine-generated prompt. When `source` is present, please accept only `"user"` while continuing to accept an absent source for external sessions.

app-server/src/agents/agentAdapter.ts
```diff
@@ -449,6 +474,66 @@ export const resolveProviderExecutable = (
...
+    for (const [open, close] of WRAPPING_QUOTES) {
+      if (title.length >= 2 && title.startsWith(open) && title.endsWith(close)) {
+        title = title.slice(1, -1).trim()
+        stripped = true
+      }
+    }
+  }
+  title = title.replace(/[.,;:!?…]+$/, "").trim()
+  if (title.length > 50) {
+    const cut = title.slice(0, 50)
```
jeremytondo, line 537: **[P3] Truncate on Unicode boundaries.** `slice(0, 50)` counts UTF-16 code units and can leave a lone surrogate when an unbroken title crosses the limit near an emoji—for example, `"a".repeat(49) + "😀"` retains only `\ud83d`, which later renders as `�`. Please truncate by code points or grapheme clusters before applying the word-boundary rule.

jeremytondo (commented):
Two actionable edge cases from the full review:

PR #159, opened 2026-08-11
Cmd-digit jump shortcuts for sidebar threads and terminals (ATC-165)

jeremytondo:
Implements [ATC-165](https://linear.app/elevenideas/issue/ATC-165/cmd-shortcuts-for-threads-and-terminals): while exact ⌘ is held, sidebar thread cards swap their connection name + status dot for a ⌘N badge (same trailing slot as the hover Pin/Archive chips, so cards never resize) and ⌘1–9 jumps to that thread. Exact ⌥⌘ does the same for the active project's standalone terminals (⌥⌘1–9, badge at the far right of the row).

## How it works

- **`SidebarShortcuts` (new)** is the single ordering authority feeding both the badges and the dispatch: threads number 1–9 over what the collapse state reveals (visible pinned rows first, then visible recent rows; rows behind "N more" get no number), terminals over the expanded Terminals section, capped at nine. Both shortcut sets keep working while the sidebar is hidden (⌘B), using the same ordering.
- **`WindowKeyboardRouter`** intercepts exact ⌘digit / ⌥⌘digit ahead of keymap lookup in the idle state: repeats are consumed without redispatch, the command-palette suspension forwards, and a pending leader sequence keeps sequence semantics. The combos are always consumed so an unnumbered digit never falls through to the focused terminal.
- **`KeyboardMonitorHost`** adds a `.flagsChanged` local monitor publishing window-scoped held modifiers on the router; `didResignKey`/`didResignActive` reset them (⌘Tab-ing away never leaves badges stuck on) and `didBecomeKey` reseeds from `NSEvent.modifierFlags` so returning with ⌘ still held lights badges immediately.
- **`Keymap`** reserves ⌘1–9 and ⌥⌘1–9 as protected triggers: a user binding one in `[keybindings]` gets a conflict diagnostic instead of a silent collision. Not rebindable, not in the command palette.

## Notes

- Caps Lock and Fn deliberately don't break the exact-modifier match (they're latched/hardware states, and the whole keymap already normalizes them away).
- Digit matching is character-based like the rest of the KeyStroke system, so layouts where digits require Shift (AZERTY) don't get these shortcuts — consistent with existing keymap behavior.

## Testing

- New `SidebarShortcutsTests`: ordering across collapse states, the 9 cap, archived-filter behavior, terminal gating, stroke matching, and dispatch through a real `AppModel`/`WindowState`.
- `KeyboardRouterTests`: exact-modifier dispatch, repeat/suspension/pending-sequence handling, and modifier reset on resign + reseed on become-key against a real `NSWindow`.
- `KeymapResolutionTests`: reservation diagnostics for both modifier sets.
- Full macOS suite: 226 tests in 25 suites, all passing. Codex (Sol, high reasoning) review findings addressed (become-key reseed; isolation diagnostic).

https://claude.ai/code/session_01J9E8hwWZwWHz3nDFFrdHJN

### DIFF
--- a/macos/ATC/SidebarShortcuts.swift
+++ b/macos/ATC/SidebarShortcuts.swift
@@ -1,0 +1,103 @@
+// The hardcoded sidebar jump shortcuts (⌘1–9 threads, ⌥⌘1–9 terminals):
+// one ordering authority that feeds both the badges the sidebar renders and
+// the dispatch the keyboard router performs, so display and behavior cannot
+// drift.
+//
+// Numbering is what the collapse state reveals, in render order: visible
+// pinned rows first, then visible recent rows, capped at nine; terminals are
+// the active project's standalone terminals, gated on the section being
+// expanded. The shortcuts deliberately do not consult sidebar visibility —
+// muscle memory keeps working while the sidebar is hidden (⌘B), using the
+// exact ordering the sidebar would show.
+
+import ATCAppServerAPI
+import Foundation
+
+enum SidebarJump: Equatable, Sendable {
+    /// Zero-based slot: ⌘1 is `.thread(slot: 0)`.
+    case thread(slot: Int)
+    case terminal(slot: Int)
+}
+
+enum SidebarShortcuts {
+    /// Rows a sidebar section shows before its More control takes over.
+    static let initialRowLimit = 5
+    /// ⌘1…⌘9 — there is no ⌘0 and nothing past nine.
+    static let slotCount = 9
+
+    /// The jump a stroke maps to, on an exact modifier match only: ⌘ alone
+    /// targets threads, ⌥⌘ targets terminals, anything else is no jump.
+    static func jump(for stroke: KeyStroke) -> SidebarJump? {
+        guard let digit = Int(stroke.key), (1...9).contains(digit) else { return nil }
+        if stroke.modifiers == [.command] { return .thread(slot: digit - 1) }
+        if stroke.modifiers == [.command, .option] { return .terminal(slot: digit - 1) }
+        return nil
+    }
+
+    static func limited<T>(_ items: [T], expanded: Bool) -> [T] {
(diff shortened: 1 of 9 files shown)


macos/ATC/SidebarShortcuts.swift
```diff
@@ -0,0 +1,105 @@
...
+        isSectionExpanded: Bool
+    ) -> [Terminal] {
+        guard isSectionExpanded else { return [] }
+        return Array(terminals.prefix(slotCount))
+    }
+
+    // Nonisolated: pure formatting, usable from nonisolated closure contexts
+    // such as `Optional.map` without an isolation diagnostic.
+    nonisolated static func threadBadgeLabel(_ number: Int) -> String {
+        KeyStroke(key: "\(number)", modifiers: [.command]).displayDescription
```
jeremytondo, line 67: `nonisolated` here only moves the isolation diagnostic into this body: the macOS build now warns that `KeyStroke.displayDescription` and `Modifiers.command`/`.option` are main-actor isolated (the same applies to `terminalBadgeLabel` below). These become errors under Swift 6. Could we keep these helpers genuinely nonisolated—for example, format the two fixed glyph prefixes directly—or explicitly make the underlying pure `KeyStroke` formatting API nonisolated?

jeremytondo (commented):
Found one concurrency-isolation issue in the new badge formatting helpers.

macos/ATC/SidebarShortcuts.swift
jeremytondo: Agreed — the `nonisolated` on these helpers only relocated the diagnostic, since `KeyStroke` picks up the target's MainActor default isolation. Fixed in 8c2d1bb by taking the second option: `KeyStroke` (plus `TriggerError`/`ParsedKeySequence`) is now a `nonisolated` type, keeping one formatting authority instead of hardcoding the glyph prefixes here. Build is clean of isolation diagnostics and all 226 macOS tests pass.

PR #160, opened 2026-08-11
macOS: keyboard-first New Thread launcher (ATC-166)

jeremytondo:
Rebuilds the New Thread sheet as a compact, keyboard-first launcher per [ATC-166](https://linear.app/elevenideas/issue/ATC-166/quick-keyboard-shortcut-for-selecting-an-agent-when-starting-a-thread). All client-side in `macos/ATC`; no server or contract changes.

## What changed

- **Search-first layout** on `SheetScaffold` (non-form): project search field, fuzzy-filtered results (shared `QueryMatcher`, moved to `Shared/`), always-visible Agent and Directory rows, Cancel/Create.
- **Keyboard**: ↑/↓/Ctrl-N/Ctrl-P move the highlight while search keeps focus; ⌘1/⌘2 select agents (never create; beep with cooldown on unavailable); Shift-⌘-G focuses the directory field; Return creates, Escape cancels.
- **Ordering**: empty query puts the contextual project first, rest alphabetical; with a query, project-name matches lead and connection-name/directory matches follow. Matching uses the bare project name so duplicate-name labels' `· Connection` suffixes never rank as primary matches.
- **Agent selection**: last-used agent persists via `AppStorage`, preserved across project changes when available, else adopts the first available; unavailable agents render disabled with the server's reason.
- **Directory**: highlight changes reset the field to the new project's default (discarding overrides); existing debounced server-side validation gates Create.
- **Name field removed** — threads auto-name after their first prompt (ATC-155).

New pure-logic `NewThreadLauncherModel` carries filtering/ordering, wraparound navigation, and the agent preserve/adopt + last-used rules; `NewThreadLauncherModelTests` covers them (240 tests pass).

## Verification

- `mise run macos:test` — 240 tests, all green.
- Manual smoke in a built app: ⌘N opens with search focused, typing filters, ⌘2 moves agent selection without stealing focus, no-match empty state renders, sheet blocks sidebar ⌘-digit jumps (key-window monitor guard).
- Codex (Sol, high reasoning) full review: no blockers; fixed duplicate-name match promotion and VoiceOver shortcut discoverability from its findings.

https://claude.ai/code/session_01Hbgm3wtYYCzKktremM7VBy


macos/ATC/Features/Threads/NewThreadSheet.swift
```diff
@@ -122,46 +338,65 @@ struct NewThreadSheet: View {
...
-              let available = agents.first(where: \.available)
-        else { return }
-        agentID = available.id
+    /// Changing the highlighted Project resets the directory to its default,
+    /// discarding any override — the field always shows where the new thread
+    /// would run.
+    private func selectProject(_ ref: ProjectRef?, in rows: [NewThreadLauncherModel.Row]) {
+        guard selectedProject != ref else { return }
+        selectedProject = ref
+        workingDirectory = rows.first { $0.id == ref }?
```
jeremytondo, line 365: Invalidate the directory verdict when the selected project changes. After the previous directory has reached `.checked(.available)`, this assignment changes the path (and sometimes the server) while `directoryState` remains available. `WorkingDirectoryField.check()` waits 300 ms before setting `.checking`, so Return remains enabled during that window and submits a path/server pair that has not been validated. The same stale-verdict window exists while manually editing the field. Please reset validation immediately when the `CheckKey` changes (or carry the checked key in the state and require it to equal the current path/connection) before allowing Create.

macos/ATC/Features/Threads/NewThreadSheet.swift
```diff
@@ -1,100 +1,316 @@
...
+                lastUsedRawValue: lastAgentRaw,
+                agents: agents
+            )
+            if selectedProject == nil {
+                selectProject(rows.first?.id, in: rows)
+            }
+        }
+        // Detection is per Connection; re-run it whenever the highlighted
+        // Project moves to another one.
+        .task(id: selectedProject?.connectionID) {
```
jeremytondo, line 78: Avoid tying this refresh directly to the cancellable `.task(id:)`. Moving the highlight to a project on another connection cancels the previous task; `AgentsStore.refresh()` routes that cancellation through `RefreshState.run`, whose catch records it as `lastError`. Because the next task refreshes a different per-connection store, the connection just left can retain a synthetic agents failure and later settle as unreachable. Either launch an on-demand refresh whose lifetime is not cancelled by highlight changes, or make `RefreshState` ignore cancellation before recording a failure.

jeremytondo (commented):
The launcher direction looks solid and the full macOS suite passes, but two state-lifecycle races can bypass validation or leave connection state poisoned during normal keyboard navigation.

macos/ATC/Features/Threads/NewThreadSheet.swift
jeremytondo: Fixed in 4ec7256: `WorkingDirectoryField.check()` now drops the state to `.checking` synchronously as soon as the (path, connection) key changes, before the 300 ms debounce — so `isAvailable` (and Create) go false immediately on a highlight change or manual edit, for every sheet that uses the field.

macos/ATC/Features/Threads/NewThreadSheet.swift
jeremytondo: Fixed in 4ec7256, via the second option: `RefreshState.run` now returns without recording when the error is a `CancellationError` or the surrounding task is cancelled (covers transports that surface cancellation as their own error type). Keeps the launcher on structured `.task(id:)` concurrency and protects every store from any cancelled caller. Added `RefreshStateTests` covering both cancellation shapes and recovery.

PR #219, opened 2026-08-18
Chat mode model settings: contract, migration, adapters, runtime (ATC-210)

jeremytondo:
Server half of ATC-205 (Chat mode model settings) — [ATC-210](https://linear.app/elevenideas/issue/ATC-210). The macOS half (#220, ATC-211) is stacked on this branch; **merge this first, then that** — a wire-contract change turns macOS CI red on a server-only PR (`Thread.settings` is required), which is expected here.

## What
- **Contract**: `Thread.settings { model, reasoning?, mode, access }`, `UpdateThreadRequest.settings` patch, `Agent.defaults` (write-through per-agent defaults a new thread inherits), `GET /agents/:agentId/models` (server-cached catalog), `InvalidThreadSettings` (400). `openapi.json` regenerated.
- **Migration** `0007_thread_settings`: four columns on `threads` (legacy rows backfilled), `agent_defaults` table.
- **Seam** (`agentAdapter.ts`): `startTurn(input, settings)` / `createSession` / `resumeSession` carry settings; adapters push only what differs from the live session; `settings` events (partial provider reports) on writer + observation feeds; `listModels`.
- **Codex**: `thread/start` carries model + ladder; `turn/start` pushes the diff vs the reply echo plus `approvalsReviewer: "user"` and `collaborationMode` (probed live: it requires `initialize` with `experimentalApi: true` — T3Code opens the same way — and Codex answers every override with `thread/settings/updated`, which becomes the settings event); paginated `model/list`.
- **Claude**: the child spawns with model / effort / permission mode (+ `allowDangerouslySkipPermissions` so Full access can be switched to mid-session); resident turns push via `setModel` / `applyFlagSettings` / `setPermissionMode`; each `system/init` reports the applied model (mapped back through the child's catalog) and permission mode, effort via the undeclared `getSettings()` behind a narrow interface (warns once on decode failure); catalog via a prompt-less query.
- **Runtime**: every turn starts with the stored settings (re-read under the start lock); provider reports are adopted (provider state wins) except fields that merely confirm the writer's own push — a delayed confirmation (Claude's spawn takes seconds) must not roll back a change made meanwhile.
- **Claude catalog names carry the version**: the CLI names models by family alone (`Fable`, `Opus (1M context)`); the version is folded in from the wire id (`claude-fable-5[1m]` → `Fable 5`, `Opus 5 (1M context)`, `Haiku 4.5`) so the composer's model control says which generation. A name the id does not explain is left as is.
- **Threads.update**: validates a model/reasoning change against the catalog before writing anything (only those fields consult the provider); every settings change writes through to the agent's defaults, in one transaction with the row. Settings writes are a compare-and-swap (`setSettingsIfUnchanged`): a client patch and a provider report — or two disjoint patches — that read the same baseline re-apply themselves on top instead of overwriting each other; a patch that resolves to what the row already holds is a no-op (no write, no write-through, no event).

## Deviations from the issue text (deliberate)
- **Claude Auto = `permissionMode: "auto"`**, not `dontAsk`. Probed live: `dontAsk` denies Write and Bash outright with no rules configured (worse than Supervised); `auto` (the classifier mode, T3Code's choice too) works within the workspace unasked and routes borderline calls to `canUseTool`. A model without auto support (Haiku) silently runs `default`; the init report adopts that as Supervised.
- `AgentModel` has no `supportsEffort` — `supportedEffortLevels` being empty says the same thing.
- On a model change the current reasoning level carries over when the new model supports it, else the model's default applies (the acceptance criterion; the "always reset" wording would drop a user's xhigh going Fable → Sonnet).
- Legacy rows are backfilled by the migration (seed model per agent + `high`); provider state corrects a stale model on the next attach.

## Verified
- `mise run -C app-server check`, `mise run contract:check` (ATCKit fixture updated here — the one shared-contract test).
- Live: `ATC_SMOKE=1` Claude adapter + thread runtime smoke pass with the new spawn options; Codex `turn/start` overrides / `collaborationMode` / `thread/settings/updated` / `model/list` probed against the live shared server (scratch thread, archived + deleted after).

https://claude.ai/code/session_012Z59VLDXmduFvKXowv3uD1

## Summary by CodeRabbit

* **New Features**
  * Added thread settings for model, reasoning level, mode, and access.
  * Settings can be updated per thread and saved as defaults for new threads.
  * Added agent model catalogs with supported models and reasoning levels.
  * Settings apply when starting, resuming, and continuing conversations, including live updates.
* **Bug Fixes**
  * Added validation for unsupported models and reasoning levels.
  * Added clear errors for invalid settings and unavailable providers.
* **Documentation**
  * Updated the API contract for thread settings, model catalogs, and related responses.

### DIFF
--- a/app-server/src/agents/claudeAdapter.ts
+++ b/app-server/src/agents/claudeAdapter.ts
@@ -1,8 +1,11 @@
 import type {
   CanUseTool,
+  EffortLevel,
   HookCallbackMatcher,
   HookEvent,
+  ModelInfo,
   Options,
+  PermissionMode,
   PermissionResult,
   PermissionUpdate,
   SDKMessage,
@@ -30,12 +33,16 @@ import type {
   AgentConnection,
   AgentEvent,
   AgentIdentityMismatch,
+  AgentModel,
   AgentProtocolError,
   AgentSessionEvent,
   AgentUnavailable,
   HistoryTurn,
+  ProviderSettings,
+  ThreadAccess,
   ThreadItem,
   ThreadRequest,
+  ThreadSettings,
 } from "./agentAdapter.ts"
 import {
   AgentResumeFailed,
@@ -51,6 +58,7 @@ import {
   withToolStatus,
 } from "./agentAdapter.ts"
 import * as path from "node:path"
+import { isReasoningLevel } from "../api/contract.ts"
 import * as ClaudeHooks from "./claudeHooks.ts"
 import * as ClaudeItems from "./claudeItems.ts"
 import { AppConfig } from "../platform/config.ts"
@@ -136,6 +144,28 @@ import * as Subprocess from "../platform/subprocess.ts"
 //     TUI's submitted prompt as an observed `userMessage` item (the seam's
 //     one hooks-carried item), so a Chat reader shows what the TUI is
 //     working on before the turn's re-read lands.
+//   - Settings (ATC-205): a fresh child is spawned with the caller's model,
+//     effort, and permission mode, so its first turn pushes nothing; later
+//     turns on the resident connection push only what changed, through the
+//     live control requests (`setModel`, `applyFlagSettings({effortLevel})`,
+//     `setPermissionMode`) before the input goes out. The ladder is ONE knob
+//     here — Supervised `default`, Auto-accept edits `acceptEdits`, Auto
+//     `auto` (the classifier mode: works within the workspace unasked and
+//     still routes borderline calls to canUseTool; `dontAsk` denies every
+//     unlisted tool outright, probed 2026-08-18), Full access
+//     `bypassPermissions` — and Plan mode is `permissionMode: "plan"`, which
+//     stands in for the access rung while planning. The child is always
+//     spawned with `allowDangerouslySkipPermissions` so Full access can be
+//     switched to mid-session; the mode itself is only ever what the caller
+//     asked for. Every `system/init` (one per turn) reports the model and
+//     permission mode the child actually applied — a model without auto
+//     support silently runs `default` — and `getSettings().applied.effort`
+//     the effort; both come back as the seam's `settings` event, the model
+//     mapped through the child's own catalog to the alias the catalog lists
+//     (init reports the wire id: `sonnet` → `claude-sonnet-5`). getSettings
+//     is undeclared on the SDK's Query type (0.3.235) but real: it is reached
+//     through a narrow local interface and schema-decoded; a decode failure
+//     keeps the caller's stored effort and warns once.
 
 /** The Claude Code version this adapter was validated against. */
 const CLAUDE_TESTED_VERSION = "2.1.235"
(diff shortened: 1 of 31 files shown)


app-server/src/agents/claudeAdapter.ts
```diff
@@ -144,11 +174,108 @@ export class ClaudeAdapter extends Context.Service<ClaudeAdapter, AgentAdapter>(
...
+        displayName: model.displayName,
+        description: model.description,
+        isDefault:
+          defaultWire !== undefined && (model.resolvedModel ?? model.value) === defaultWire,
+        supportedEffortLevels: levels,
+      }
+      // The CLI applies `high` unless told otherwise (probed 2026-08-18).
+      if (levels.length > 0) entry.defaultEffortLevel = "high"
+      return entry
+    })
```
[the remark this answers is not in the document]
jeremytondo: Fixed in dd55b39: the default is now clamped to what the model lists (`high` when listed, else the first level). Validated: `normalizeClaudeCatalog` test + `mise run -C app-server check`.

app-server/src/agents/claudeAdapter.ts
```diff
@@ -938,18 +1095,112 @@ export const layerWith = (adapterOptions: ClaudeAdapterOptions) =>
...
+            yield* control("setModel", () => handle.setModel(settings.model))
+            session.live = { ...session.live, model: settings.model }
+          }
+          const effort = claudeEffort(settings.reasoning)
+          if (effort !== undefined && session.live.reasoning !== effort) {
+            yield* control("applyFlagSettings", () =>
+              handle.applyFlagSettings({ effortLevel: effort }),
+            )
+            session.live = { ...session.live, reasoning: effort }
+          }
```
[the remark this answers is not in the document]
jeremytondo: Claude half fixed in dd55b39: `pushSettings` now treats "no reasoning" as `effortLevel: null` and clears the flag layer (and the baseline) when the live effort differs — covered by the new step in the resident-connection test (`setModel:haiku`, `applyFlagSettings:null`). Codex half: won't fix — every Codex model lists effort levels (`reasoning` is only undefined for a model with no effort support), and `effort: null` on `turn/start` means "no override", not "clear".

app-server/src/threads/threadRuntime.ts
```diff
@@ -630,6 +647,29 @@ const make = (options: ThreadRuntimeOptions) =>
...
+      pushed: ThreadSettings | undefined,
+    ): Effect.Effect<void> =>
+      Effect.gen(function* () {
+        const current = yield* repository.get(record.id)
+        if (Option.isNone(current)) return
+        const adopted = applyProviderSettings(current.value.settings, reported, pushed)
+        if (adopted === undefined) return
+        const updated = yield* repository.setSettings(record.id, adopted)
+        if (Option.isNone(updated)) return
+        yield* events.publish({ resource: "thread", id: record.id, change: "updated" })
```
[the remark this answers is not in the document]
jeremytondo: Won't fix, deliberately: the issue defines the write-through as the settings of the last thread a user CHANGED ("Changing any thread's setting writes it … there is no second writer to drift"). A provider report is not a choice — the common Claude case is the CLI silently downgrading `auto` to `default` on a model without auto support, and writing that through would flip every new Claude thread to Supervised. Documented in the module header and the Linear thread.

app-server/src/threads/threads.ts
```diff
@@ -663,20 +731,31 @@ export const layerWith = (options: ThreadsOptions) =>
...
+            // including updatedAt (the same rule the other repositories
+            // apply).
+            if (patch.name === undefined && settings === undefined) {
+              return yield* assembleAlone(record)
+            }
+            const renamed =
+              patch.name === undefined ? record : yield* repository.rename(id, patch.name)
+            const settled =
+              settings === undefined ? renamed : yield* storeSettings(renamed, settings)
+            yield* events.publish({ resource: "thread", id, change: "updated" })
```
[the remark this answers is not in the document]
jeremytondo: Fixed in dd55b39: an empty `settings` object is treated as an empty patch (nothing written, updatedAt untouched, no event) — asserted in threadSettings.test.ts.

jeremytondo:
Nitpick dispositions (round 1, head dd55b39):
- fake-codex-listener echoes the model's default effort on `thread/start` — **fixed** (the fixture now mirrors the real reply; the first-turn override assertion still holds since the desired `high` differs from fake-sol's `medium`).
- openapi.test: assert the catalog is non-empty — **fixed**.
- codexAdapter.test: assert on the last `settings` event (`findLast`), not the last sink entry — **fixed**.
- fakeClaudeQuery: snapshot the spawn options — **fixed**.
- `SANDBOX_POLICY_TYPE` typed by its keys (`as const`, `ACCESS_TO_THREAD_KNOBS.sandbox: keyof typeof …`) — **fixed**.
- Claude "read the catalog once per session, not per turn" — **won't fix**: `supportedModels()` resolves from the SDK's cached initialization result (no round-trip); the per-init read is what keeps a hot-swapped catalog truthful and costs nothing.
- `memoizeSuccess` "narrow invalidation to the driving fiber" — **won't fix / already so**: `onInterrupt` sits inside `withPermits(1)`, so only the permit holder — the fiber actually driving `cached` — can be interrupted there; waiters interrupt on the semaphore acquisition, never the memo.
- threadSettings.test "select the session by provider id" — **won't fix**: the provider session id is not on the contract; the prompt text is unique within the test.

app-server/src/agents/codexAdapter.ts
```diff
@@ -1472,6 +1672,49 @@ export const layer = Layer.effect(CodexAdapter)(
...
+              const entry: { -readonly [K in keyof AgentModel]: AgentModel[K] } = {
+                value: model.model,
+                displayName: model.displayName,
+                description: model.description,
+                isDefault: model.isDefault,
+                supportedEffortLevels: levels,
+              }
+              if (defaultLevel !== undefined) entry.defaultEffortLevel = defaultLevel
+              return entry
+            })
```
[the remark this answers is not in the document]
jeremytondo: Fixed in 98294e3: the Codex catalog's default effort is now clamped to the model's own listed levels (`isReasoningLevel(reported) && levels.includes(reported) ? reported : levels[0]`), matching the Claude side. Covered by a new `fake-terra` fixture model that reports `defaultReasoningEffort: "high"` with only `low`/`medium` listed — `listModels` decodes it as `defaultEffortLevel: "low"`. `codexAdapter` + `threads` tests, typecheck, lint pass.

app-server/test/fixtures/fake-codex-listener.ts
```diff
@@ -304,8 +416,36 @@ const handle = (
...
+      if (typeof params["sandboxType"] === "string")
+        thread.settings.sandboxType = params["sandboxType"]
+      if (typeof params["mode"] === "string") thread.settings.mode = params["mode"]
+      respond({})
+      return broadcast("thread/settings/updated", settingsUpdatedParams(thread))
+    }
+    case "model/list": {
+      // Paginated like the real reply: one model per page under
+      // FAKE_CODEX_MODEL_PAGED, else everything at once.
+      if (process.env["FAKE_CODEX_MODEL_PAGED"] !== "1") {
```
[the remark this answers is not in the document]
jeremytondo: Won't fix. `fake-codex-listener.ts` is a test fixture that runs as its own child process standing in for the external `codex app-server` binary — it is not server code and has no `AppConfig` (the tests spawn it through `ATC_CODEX_EXECUTABLE` and configure it the way any external binary is configured: environment variables, see the `FAKE_CODEX_*` list in the header). The `AppConfig` rule guards `src/`; this switch also predates this PR (only the model list gained an entry here).

app-server/src/threads/threads.ts
```diff
@@ -663,20 +731,31 @@ export const layerWith = (options: ThreadsOptions) =>
...
+            // Validated; now the writes. An empty patch changes nothing —
+            // including updatedAt (the same rule the other repositories
+            // apply).
+            if (patch.name === undefined && settings === undefined) {
+              return yield* assembleAlone(record)
+            }
+            const renamed =
+              patch.name === undefined ? record : yield* repository.rename(id, patch.name)
+            const settled =
+              settings === undefined ? renamed : yield* storeSettings(renamed, settings)
```
jeremytondo: **[P1] Treat effective no-op settings patches as no-ops.**

Only an empty object is skipped above. A patch such as `{ mode: "chat" }` on an already-chat thread resolves to the same full settings but still reaches `storeSettings`, updates `updatedAt`, publishes an event, and—more importantly—writes this thread's entire settings into the agent defaults. In the macOS PR, checked menu entries remain actionable, so reselecting a checked value on an older thread can silently reset new-thread defaults even though nothing changed.

Please compare the resolved settings with `record.settings` and skip the settings write, defaults write, and event when they are identical.

app-server/src/threads/threadRuntime.ts
```diff
@@ -630,6 +647,29 @@ const make = (options: ThreadRuntimeOptions) =>
...
+      record: ThreadRecord,
+      reported: ProviderSettings,
+      pushed: ThreadSettings | undefined,
+    ): Effect.Effect<void> =>
+      Effect.gen(function* () {
+        const current = yield* repository.get(record.id)
+        if (Option.isNone(current)) return
+        const adopted = applyProviderSettings(current.value.settings, reported, pushed)
+        if (adopted === undefined) return
+        const updated = yield* repository.setSettings(record.id, adopted)
```
jeremytondo: **[P1] Serialize the settings read–modify–write across API patches and provider adoption.**

This path reads the current full settings and later replaces all four fields. `Threads.update` independently does the same. Two disjoint partial patches—or a provider report racing a client patch—can both read the same baseline, then the last full-row write silently erases the other change. That violates partial-patch semantics and the provider-adoption contract.

Both callers need one shared per-thread mutation seam, or an atomic repository operation/conditional merge, covering the read, merge, and write.

app-server/src/threads/threads.ts
```diff
@@ -617,6 +634,57 @@ export const layerWith = (options: ThreadsOptions) =>
...
+        record: ThreadRecord,
+        settings: ThreadSettings,
+      ): Effect.Effect<ThreadRecord, ThreadNotFound> =>
+        Effect.gen(function* () {
+          const updated = yield* repository.setSettings(record.id, settings)
+          // A delete racing the write: the endpoint's typed 404.
+          if (Option.isNone(updated)) {
+            return yield* Effect.fail(new ThreadNotFound({ threadId: record.id }))
+          }
+          if (isAgentId(record.agentId)) yield* registry.setDefaults(record.agentId, settings)
```
jeremytondo: **[P2] Persist the thread settings and write-through defaults atomically.**

The thread row is committed first and the defaults row is upserted afterward. Interruption, process failure, or a database defect between those effects leaves the endpoint failed with the thread changed and `agent_defaults` stale, breaking the stated write-through invariant.

Please put the two writes in one transaction, ideally behind a repository seam that owns this cross-table invariant.

app-server/src/threads/threadRuntime.ts
```diff
@@ -1265,7 +1320,11 @@ const make = (options: ThreadRuntimeOptions) =>
         if (providerSessionId === undefined) return yield* markInterrupted(record, turn)
         const scope = Scope.forkUnsafe(serviceScope)
         const connection = yield* adapter
-          .resumeSession({ providerSessionId, cwd: record.workingDirectory })
+          .resumeSession({
+            providerSessionId,
+            cwd: record.workingDirectory,
+            settings: record.settings,
```
jeremytondo: **[P2] Restore the running turn's pushed-settings baseline on reattach.**

Although `settings` are passed into `resumeSession`, the `Writer` created immediately below still has `pushed: undefined`. After a restart, any settings notification from the already-running turn is therefore treated as an external provider change and may roll back settings the user stored for the next turn.

The robust fix is to persist the actual settings used when the running turn started and restore them into `writer.pushed` during reattachment. Using the current thread record alone is not sufficient when settings changed mid-turn before the restart.

jeremytondo (commented):
Full correctness and simplicity review. The provider mappings, catalog normalization, migration shape, and test coverage look thoughtful. I found four settings-consistency issues that should be addressed before merge.

app-server/src/threads/threads.ts
jeremytondo: Fixed in ba24490. `Threads.update` now resolves the patch and compares it with the row (`sameSettings`); an identical result is a no-op — no row write (updatedAt unchanged), no defaults write-through, no event. Covered in `threadSettings.test.ts` ("a no-op patch writes nothing": re-picking a thread's own model+mode leaves `updatedAt` untouched and the agent defaults at another thread's later values). The macOS side additionally skips sending a patch when the checked entry is re-picked.

app-server/src/threads/threadRuntime.ts
jeremytondo: Fixed in ba24490 with an atomic conditional write rather than a shared lock: `ThreadRepository.setSettingsIfUnchanged(id, expected, next)` guards the UPDATE on all four settings columns (null-safe on reasoning, the `renameIfUnchanged` precedent). `Threads.update` reads → resolves (catalog) → CAS, and on losing the race re-reads and re-applies the patch on top; the runtime's `adoptSettings` does the same read → merge → CAS. Two disjoint patches, or a provider report racing a client patch, therefore compose instead of the last full-row write winning. Test: the model patch's catalog read is parked (`fakeAgent.gateModels`) while a mode patch lands; the model patch then completes with both changes standing (without the CAS the mode change is lost).

app-server/src/threads/threads.ts
jeremytondo: Fixed in ba24490: the settings write and the defaults write-through run inside one `sql.withTransaction` (`Threads` holds the `SqlClient` for that alone — no SQL leaves the repositories; persistence.ts's header notes the allowance). Both repositories share the client, so the transaction connection propagates through the fiber to both statements.

app-server/src/threads/threadRuntime.ts
jeremytondo: Won't fix (documented instead, ba24490). Restoring the running turn's pushed baseline needs the settings each turn started with persisted per turn (a turns column + repository + runtime plumbing) for a window that is a server restart *inside* a running turn on a shared provider server *after* a mid-turn settings change; setting `pushed` from the current row does not help (in that case the row already differs from what the turn runs, so the report is adopted either way). Provider-state-wins makes the outcome defensible — the row briefly follows the running turn again — and the next explicit change corrects it. The reattach path now carries a comment stating exactly this trade-off, so it is a known gap rather than a hidden one; happy to add the per-turn persistence if you want it.

app-server/src/threads/threadSettings.ts
```diff
@@ -0,0 +1,129 @@
...
+  readonly model: string
+  readonly reasoning: string | null
+  readonly mode: string
+  readonly access: string
+}): ThreadSettings => ({
+  model: row.model,
+  ...(isReasoningLevel(row.reasoning) ? { reasoning: row.reasoning } : {}),
+  mode: isMode(row.mode) ? row.mode : "chat",
+  access: isAccess(row.access) ? row.access : "auto",
+})
```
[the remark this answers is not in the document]
jeremytondo: Good catch — fixed in b9dc4e3. `setSettingsIfUnchanged` now reads the row first, compares the caller's `expected` against the *decoded* row (`sameSettings(settingsFromColumns(stored), expected)`), and guards the UPDATE on the literals *as stored* (`stored.mode`, `stored.access`, …), so a coerced row matches and the write normalizes it. Test added: a row poked to `mode = 'from-the-future'` reads as `chat`, still takes an access patch, and afterwards holds `chat` in the column.

jeremytondo:
Folded into #220: a wire-contract change and its client can only be green together (main requires both `app-server-ci-ok` and `macos-ci-ok`), so the stack cannot merge in order. #220 now carries this commit under the macOS one; every review thread here was addressed and resolved on the same commit.

PR #243, opened 2026-08-24
ATC-249 | Restore reference checkout task

jeremytondo:
## Summary

- restore `mise run refs` for read-only local reference checkouts
- fetch T3 Code, Herdr, the Agent Client Protocol specification, and zmx v0.6.0
- document the task while keeping archived Effect/App Server tooling absent

## Testing

- `mise tasks validate`
- `mise task info refs`
- `mise run --dry-run --skip-tools refs`
- `mise run check`
- `git diff --check`

## Dependency

This is a stacked PR targeting #242 because ATC-249 builds on the root `mise.toml` introduced by ATC-248.


jeremytondo:
@coderabbitai review

PR #244, opened 2026-08-25
ATC-259 | Server chassis: authenticated /v1 listener, tailnet exposure, config, and Cobra CLI

jeremytondo:
Implements the Foundation server chassis per the ATC-250 charter and the ATC-246/ATC-247 decided records: [ATC-259](https://linear.app/elevenideas/issue/ATC-259/server-chassis-authenticated-v1-listener-tailnet-exposure-and-config).

## What's here (one commit per step)

1. **Framework spike** — stdlib `net/http` vs Huma v2 conformance experiment under `experiments/http-framework/`, judged on the real first routes (health, auth middleware, version headers).
2. **Server chassis** — `atc server run`: authed `/v1` listener on `127.0.0.1:7331`, bearer auth on every request **including loopback** (the one named deviation from legacy), health endpoint, version headers both ways, token store with legacy-proven mechanics (`atc_<43 base64url>`, 0600 atomic writes, re-read per verify), `atc server token [rotate]`, JSON-lines log at `~/.local/state/atc/atc.log`, and the fresh-designed config module: flags > `ATC_<KEY>` env > `~/.config/atc/config.toml` > defaults, snake_case keys, unknown keys refuse startup.
3. **Tailnet exposure** — `--tailscale` flag / `tailscale = true` config key supervising a loopback-fronting `tailscale serve` for the server's lifetime (legacy-proven default per the charter).
4. **Cobra CLI** — the hand-rolled dispatch replaced with a Cobra command tree; same commands and behavior, `rotate` promoted to a real subcommand, generated help and shell completion. Config deliberately stays hand-rolled (go-toml/v2 + explicit env overlay); decision trail and the koanf revisit trigger are recorded on the Linear issue.

## Testing

- `mise run check` (build, gofmt, vet, `go test -race ./...`) passes; CLI tests drive the tree through the same `run()` seam and needed no changes for the Cobra migration.
- Smoke-tested the built binary against an isolated XDG dir: help, token mint/rotate, bad-invocation rejection, clean SIGINT shutdown.


cmd/atc/main.go
```diff
@@ -23,58 +36,218 @@ func main() {
...
+	}
+	// Flags outrank the loaded config, but only when the user set them.
+	flags := cmd.Flags()
+	if flags.Changed("port") {
+		if cfg.Port, err = flags.GetInt("port"); err != nil {
+			return err
+		}
+	}
+	if flags.Changed("bind") {
+		if cfg.Bind, err = flags.GetString("bind"); err != nil {
```
jeremytondo: Revalidate after applying flags. `config.Load` rejects an empty bind, but this override happens after that validation. `atc server run --bind= --port=0` currently starts on `[::]:<port>` (confirmed locally), exposing the listener on every interface instead of refusing the empty value. Invalid port flags bypass the range check for the same reason. Please centralize validation and run it after the final precedence layer.

internal/authtoken/authtoken.go
```diff
@@ -0,0 +1,165 @@
...
+		if err := os.Chmod(s.Path, 0o600); err != nil {
+			return "", err
+		}
+		return existing, nil
+	}
+	if err := os.MkdirAll(filepath.Dir(s.Path), 0o700); err != nil {
+		return "", err
+	}
+	token := generate()
+	f, err := os.OpenFile(s.Path, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0o600)
```
jeremytondo: Keep the final token path invisible until the token is complete. `O_EXCL` makes creator election atomic, but it publishes an empty file before `WriteString`; a concurrent `Ensure` can hit `ErrExist`, read during that window, and fail as malformed. A crash in the same window also leaves a permanently malformed token file. This does not meet the ATC-259 atomic-write contract. Please fully write a temporary file before atomically installing it (with coordination/no-replace semantics for competing creators), or otherwise ensure losers retry without observing partial contents.

internal/tailscale/tailscale.go
```diff
@@ -0,0 +1,288 @@
...
+	var readers sync.WaitGroup
+	watch := func(r io.Reader, keep bool) {
+		defer readers.Done()
+		scanner := bufio.NewScanner(r)
+		for scanner.Scan() {
+			line := scanner.Text()
+			if keep {
+				errTail.add(line)
+			}
+			if strings.Contains(line, "https://") {
```
jeremytondo: Treat only the expected Serve banner as readiness. Any line containing `https://` closes `ready`; current Tailscale prints HTTPS consent/login URLs when Serve or tailnet HTTPS still needs enabling, so the child can be waiting for setup while ATC logs `tailscale serving`. Matching the expected `https://<dnsName>:<port>` (or verifying `tailscale serve status`) would preserve the stated fail-closed behavior. A fake that prints an unrelated HTTPS URL before blocking would cover this.

internal/tailscale/tailscale.go
```diff
@@ -0,0 +1,288 @@
...
+	defer bannerTimeout.Stop()
+	select {
+	case <-ready:
+	case err := <-exited:
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		return fmt.Errorf("tailscale serve exited: %s", exitReason(err, &errTail))
+	case <-bannerTimeout.C:
+		cmd.Cancel()
```
jeremytondo: Bound the readiness-timeout cleanup. This calls `Cancel` manually and then waits forever for `exited`, but `Wait` is not reached until both pipe readers see EOF, and `WaitDelay` is only activated by the command context cancellation in this arrangement. If a hung child ignores SIGINT and keeps its pipes open, the supervisor never returns to its retry loop. Cancel an attempt-scoped context here (so the configured kill delay runs), or add an explicit post-interrupt kill timer before waiting for reap.

jeremytondo (commented):
Full review complete. I left four inline findings that should be addressed before merge: final config validation after flags, atomic first-token creation, precise Tailscale readiness detection, and bounded cleanup after readiness timeout. Validation run: `mise run check` and `go test -race ./...` in `experiments/http-framework` both pass.

cmd/atc/main.go
jeremytondo: Fixed in a4328dba338c. Validation now lives in an exported `config.Validate` method; `Load` still runs it, and `serverRun` reruns it after the flag layer settles, so `--bind=` and out-of-range `--port` are refused instead of reaching `net.Listen`. Covered by `TestServerRunRejectsInvalidFlagValues` (empty bind and port 70000 both error). `mise run check` passes.

internal/authtoken/authtoken.go
jeremytondo: Fixed in a4328dba338c. `Ensure` now fully writes the token to a 0600 temp file (shared `writeTemp` helper, also used by `Rotate`) and installs it with `os.Link` — atomic and no-replace, so creator election stays exclusive while the final path only ever holds a complete token; losers adopt the winner's fully-written file and a crash can no longer strand a partial credential at the token path. Covered by `TestEnsureConcurrentCallersAgree` (16 racing Ensures must agree, verify, and leave no temp files), run under `-race`. `mise run check` passes.

internal/tailscale/tailscale.go
jeremytondo: Fixed in a4328dba338c. Readiness now matches `https://<dnsName>` instead of any `https://` line, so consent/login URLs (always on login.tailscale.com, never the node's DNS name) can no longer close `ready`. One deliberate deviation from the suggested match: the port is left out of the needle because the Serve banner omits `:443` on the default HTTPS port — matching the node's own URL keeps the fail-closed property without breaking that case. Covered by the requested fake in `TestServeIgnoresUnrelatedHTTPSURLs`: an unrelated HTTPS consent URL before blocking must time out without ever logging "tailscale serving". `mise run check` passes.

internal/tailscale/tailscale.go
jeremytondo: Fixed in a4328dba338c. `serve` now runs the child under an attempt-scoped context (`context.WithCancel(ctx)`), and the readiness-timeout path cancels that context instead of calling `cmd.Cancel` directly — context cancellation is what arms exec's WaitDelay kill, so a child that ignores SIGINT and holds its pipes open is SIGKILLed after the configured delay and the supervisor returns to its retry loop. Covered by `TestServeTimeoutKillsHungChild` (fake traps INT/TERM and never prints the banner; serve must return the readiness-timeout error rather than hang). `mise run check` passes.

PR #246, opened 2026-08-25
ATC-260 | Supervisor-backed server lifecycle: start/stop/restart/status/logs/uninstall

jeremytondo:
Implements the ATC-246 lifecycle family on the ATC-259 chassis — supervisor-only, one family, no pidfile code. Closes ATC-260.

## What's here

New `internal/service` package plus the `atc server` command wiring:

- **`start`** — registration folded in: verifies `systemctl` on Linux (clear pointer to `atc server run` otherwise), pre-flight-probes the configured port when the unit isn't already supervised ("something is already serving on port N; stop it first"), re-renders the unit from `os.Executable()` + the installing shell's PATH on every start, starts via the supervisor, and health-gates (15 s / 150 ms, authenticated `GET /v1/health`). Gate timeout prints the daemon's last log lines. Idempotent: a healthy running server is left untouched (unit still re-rendered).
- **First-run notice** to stderr when no unit existed — what was registered, that it runs at every login, the `uninstall` undo, and the `loginctl enable-linger` hint on Linux. No prompt, no TTY branch.
- **`restart`** — start's path, always bouncing: `systemctl --user restart` on Linux; **bootout → wait for launchd to forget the label → bootstrap** on macOS. Amended from the ticket's original `kickstart -k` (recorded on ATC-260): bootstrap is the only launchd operation that re-reads the plist, so kickstart would bounce the process onto the stale unit and break the "never goes stale" guarantee.
- **`stop`** — supervisor stop, no disable: "stopped atc.server (still installed; returns at next login)".
- **`status`** — health probe first (source of truth; server version readable via the `Atc-Server-Version` header even on 401), unit state supplementary, client/server versions with skew flagged and a restart hint, paste-ready loopback/LAN/tailnet URLs, exit codes 0/1/2.
- **`logs`** — `journalctl --user -u atc.server` on Linux, tail of the log file on macOS, with `-f`/`-n`; unsupervised boxes get pointed at foreground `atc server run` stderr.
- **`uninstall`** — stop, remove the unit, list the config/token/(macOS) log files left behind. No purge.
- **`run`** demoted: listed last under `atc server`, labeled advanced (command sorting disabled).

## Chassis change (per ticket)

`server run` now logs slog **text** (logfmt) to stderr only and no longer opens `~/.local/state/atc/atc.log` — journal captures on Linux, the LaunchAgent's `StandardOutPath`/`StandardErrorPath` redirect captures on macOS.

## Design notes

- Lifecycle commands settle config from **file + defaults only** (no `ATC_*` env): the supervised daemon never sees shell env, so honoring `ATC_PORT` here would probe a port the daemon never serves.
- `tailscale.DNSName` exported so `status` can print the tailnet URL; the serve supervisor's preflight now wraps it.
- Boundaries per ticket: unit rendering and message formatting pure and fully tested (golden `cmp.Diff` strings, XML/systemd escaping); `launchctl`/`systemctl`/`journalctl` calls thin and untested by design.

## Testing

- `mise run check` green (build, golangci-lint, race-detector tests).
- Smoke-tested on Linux with isolated XDG state: help ordering, `status` exit codes against a live foreground server, stop/uninstall messages, the start conflict pre-flight (no unit written), journal-backed `logs`.
- Not exercised here: the real `systemctl`/`launchctl` start–stop–restart cycle (needs a login session / a Mac). Worth watching on first Mac use that bootout→bootstrap stays under the 15 s health gate.

### DIFF
--- a/cmd/atc/main.go
+++ b/cmd/atc/main.go
@@ -6,13 +6,13 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
 	"net"
 	"os"
 	"os/signal"
-	"path/filepath"
 	"runtime/debug"
 	"strconv"
 	"sync"
@@ -24,6 +24,7 @@ import (
 	"github.com/jeremytondo/atc/internal/config"
 	"github.com/jeremytondo/atc/internal/paths"
 	"github.com/jeremytondo/atc/internal/server"
+	"github.com/jeremytondo/atc/internal/service"
 	"github.com/jeremytondo/atc/internal/tailscale"
 )
 
@@ -31,6 +32,12 @@ func main() {
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 	defer stop()
 	if err := run(ctx, os.Args[1:], os.Stdout, os.Stderr); err != nil {
+		// An ExitError's report is already on stdout (e.g. `server status`
+		// exit codes); it only picks the process exit code.
+		var exit *service.ExitError
+		if errors.As(err, &exit) {
+			os.Exit(exit.Code)
+		}
 		fmt.Fprintln(os.Stderr, "atc:", err)
 		os.Exit(1)
 	}
@@ -51,6 +58,9 @@ func run(ctx context.Context, args []string, stdout, stderr io.Writer) error {
 }
 
 func newRootCmd() *cobra.Command {
+	// Help lists commands in the order they are added, so the advanced
+	// foreground `run` can sit last under `atc server`.
+	cobra.EnableCommandSorting = false
 	root := &cobra.Command{
 		Use:   "atc",
 		Short: "The ATC terminal client and server",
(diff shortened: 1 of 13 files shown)


cmd/atc/main.go
```diff
@@ -88,19 +98,134 @@ func newServerCmd() *cobra.Command {
...
+// will read: config file and defaults only. The unit stamps no ATC_*
+// environment (supervisors start services with a minimal environment), so
+// honoring the invoking shell's ATC_PORT here would probe a port the daemon
+// never serves.
+func lifecycleOptions(cmd *cobra.Command) (service.Options, error) {
+	configPath, err := paths.ConfigFile()
+	if err != nil {
+		return service.Options{}, err
+	}
+	cfg, err := config.Load(configPath, func(string) (string, bool) { return "", false })
```
jeremytondo: [P1] Keep recovery commands independent of config parsing. This helper is used by stop and uninstall, and logs calls it separately, so an unknown or malformed config key—the same condition that can prevent the daemon from booting—also blocks diagnostics, stopping, and the promised total undo. None of those three operations consumes Config. Please construct their Options without config.Load so they remain usable when config.toml is broken.

internal/service/unit.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// writeUnit renders and installs the unit for this platform. Every start
+// runs it, so the unit always reflects the current binary and PATH.
+func writeUnit(unitFile string) error {
+	executable, err := os.Executable()
+	if err != nil {
+		return err
+	}
+	pathEnv := os.Getenv("PATH")
```
jeremytondo: [P1] Propagate the XDG locations used to settle lifecycle state, or stop honoring shell-scoped XDG overrides here. UnitPath, lifecycleOptions, ensureToken, and the log path all honor XDG_CONFIG_HOME/XDG_DATA_HOME/XDG_STATE_HOME, but the supervised process receives only PATH. With an override set in the installing shell but absent from the launchd/systemd manager environment, start writes/probes one config and token while server run reads or creates the defaults; systemd may not even discover a unit written under the custom config home. The health gate then times out with a token or port mismatch. The isolated-XDG tests do not cross this supervisor boundary, so they cannot catch it.

internal/service/lifecycle.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	supervised := supervisorRunning(ctx)
+	if !supervised && portAnswering(opts.Config) {
+		// An ATC responder that is not the supervised unit is a stray
+		// foreground `atc server run`.
+		return fmt.Errorf("something is already serving on port %d (a foreground `atc server run`?); stop it first", opts.Config.Port)
+	}
+	if supervised && !bounce && probeOnce(ctx, opts, token).healthy {
+		if err := rewriteUnit(ctx, unitFile); err != nil {
```
jeremytondo: [P1] The healthy fast path does not fully perform the registration/refresh promised by start. On Linux it only daemon-reloads, so an active but disabled unit remains disabled and will not return at login. On macOS rewriteUnit only changes the plist; launchd keeps the old executable and PATH in its loaded job, so a later KeepAlive relaunch still uses stale configuration. This exposes a real conflict between leaving a healthy process untouched and the never-stale guarantee: Linux can at least enable here, while the launchd behavior needs an explicit decision or a change-aware bounce.

internal/service/lifecycle.go
```diff
@@ -0,0 +1,250 @@
...
+	_, statErr := os.Stat(unitFile)
+	firstRun := errors.Is(statErr, fs.ErrNotExist)
+	if err := writeUnit(unitFile); err != nil {
+		return err
+	}
+	if firstRun {
+		// Consent-after-the-fact (ATC-246 grill decision): no prompt, no
+		// TTY branch — identical interactive or scripted. Honest because
+		// uninstall is one cheap, total undo.
+		say(opts.Stderr, "%s", firstRunNotice(runtime.GOOS, unitFile))
```
jeremytondo: [P2] Emit this notice only after supervisor registration succeeds. At this point Linux has not daemon-reloaded or enabled the unit and macOS has not bootstrapped it. Any failure below leaves only a file on disk, yet the command has already told the user that the server is registered and now starts at every login. Retain firstRun here, but print the notice after enable/bootstrap succeeds so consent-after-the-fact describes an actual fact.

internal/service/service.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+	}
+	return runSupervisor(ctx, "launchctl", "bootstrap", domains[len(domains)-1], unitFile)
+}
+
+// launchdBootout unloads the agent from every domain; failures mean "not
+// loaded there", an expected state.
+func launchdBootout(ctx context.Context) {
+	for _, domain := range launchdDomains() {
+		exitCode(ctx, "launchctl", "bootout", domain+"/"+UnitName)
```
jeremytondo: [P1] Do not treat every bootout failure as equivalent to not loaded. Stop first confirms that the label is loaded, then calls this helper and unconditionally reports success; a permission, IPC, or transient launchctl error therefore leaves the daemon running while telling the user it stopped. Uninstall can likewise remove the plist and report success while the loaded job survives. Use the discovered loaded domain with runSupervisor so its stderr is surfaced, and ignore failures only for domains already proven absent.

internal/service/lifecycle.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	unitFile, err := UnitPath()
+	if err != nil {
+		return err
+	}
+	if runtime.GOOS == "darwin" {
+		launchdBootout(ctx)
+	} else if requireSystemctl() == nil {
+		// Non-zero here means "not loaded/enabled", an expected state.
+		exitCode(ctx, "systemctl", "--user", "disable", "--now", UnitName)
```
jeremytondo: [P1] Do not remove the unit and report success when disable --now fails. Non-zero is expected for a genuinely absent unit, but it also covers a missing user bus, permission failure, or a stop failure. In those cases this proceeds to delete the file while the process may still be running and while the enablement symlink may remain, defeating uninstall as the total undo. Distinguish not-found/not-loaded from operational failure and surface the latter before removing anything.

internal/service/logs.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/jeremytondo/atc/internal/paths"
+)
+
+// Logs shows the supervisor-captured server output: the journal on Linux, a
+// tail of the launchd-redirected log file on macOS.
+func Logs(ctx context.Context, opts Options, follow bool, lines int) error {
+	if err := supported(); err != nil {
+		return err
+	}
+	if runtime.GOOS == "linux" {
```
jeremytondo: [P2] Check whether the unit is installed before invoking journalctl. On a normal Linux machine journalctl exists even when ATC has never been supervised, so this currently prints an empty journal and succeeds instead of giving the specified explanation that a foreground server logs to stderr. The macOS branch already detects the analogous missing log file; Linux should inspect UnitPath or otherwise establish supervision first.

cmd/atc/main.go
jeremytondo: Fixed in 39193ae. Split the options builders: `lifecycleOptions` (settles config) now backs only start/restart/status; stop, logs, and uninstall use a new `recoveryOptions` that never calls `config.Load`, and `service.Options` documents that those three leave Config zero. Validated by a new `TestRecoveryCommandsSurviveBrokenConfig` (unknown-key config.toml: status surfaces the parse error; stop/uninstall/logs all proceed) plus a live smoke test with `totally_bogus = 1` in config.toml.

internal/service/unit.go
jeremytondo: Fixed in 39193ae — took the propagate option. `unitEnv()` stamps `XDG_CONFIG_HOME`/`XDG_DATA_HOME`/`XDG_STATE_HOME` into the unit alongside PATH whenever they are set in the installing shell, so the daemon resolves the same config, token, and state files the CLI probed and health-gated. This works fully on launchd (`~/Library/LaunchAgents` is XDG-independent). Known residual on Linux: the systemd user manager computes unit search paths from its own environment, so a shell-only `XDG_CONFIG_HOME` override still means the manager may not discover the unit file — but that now fails loudly at `systemctl enable` with the supervisor stderr instead of a silent health-gate timeout on mismatched token/port. Covered by `TestUnitEnvStampsXDGOverrides` and updated render goldens; ATC-260 amended (PATH was recorded as the only stamped variable).

internal/service/lifecycle.go
jeremytondo: Fixed in 39193ae with a change-aware bounce, which resolves the untouched-vs-never-stale conflict explicitly: start renders the unit first and compares it to the installed file. Healthy + unchanged → fast path that now also runs `systemctl --user enable` (repairing active-but-disabled) and skips the pointless rewrite/daemon-reload; any content change while supervised forces `bounce = true`, so Linux restarts and macOS bootout+bootstraps onto the new unit — launchd never keeps a stale loaded job past a start. Upgrade-in-place (same binary path) renders identical content, so repeated start stays a true no-op and restart remains the upgrade remedy. ATC-260 amended: \"a healthy running server is left untouched\" now reads \"…when the freshly rendered unit is unchanged\".

internal/service/lifecycle.go
jeremytondo: Fixed in 39193ae. `firstRun` is still captured from the pre-write read, but the notice now prints only after the supervisor accepted the unit (post daemon-reload/enable/start on Linux, post bootstrap on macOS) and before the health gate — so it states a fact, and still appears even if the gate later times out (registration did happen and uninstall is the undo).

internal/service/service.go
jeremytondo: Fixed in 39193ae. `launchdUnload` (the only bootout path now — stop, uninstall, and start all use it and propagate its error) skips domains proven absent via `launchctl print`, runs `bootout` through `runSupervisor` so launchctl stderr is captured, and then verifies by observation: it polls until the label is gone from every domain and returns an error carrying the bootout diagnostic if the label is still loaded at the 5s deadline. Judging by observation rather than exit status alone tolerates launchctl's exit-code quirks during wind-down while still refusing to report \"stopped\" when the job actually survives.

internal/service/lifecycle.go
jeremytondo: Fixed in 39193ae. Uninstall now distinguishes the states by checking the unit file first: when it exists, `disable --now` runs through `runSupervisor` and any failure (missing user bus, permission, stop failure) aborts the command with the supervisor's stderr before anything is removed; when it does not exist, systemctl is skipped entirely (nothing to disable — the genuinely-absent case that used to justify ignoring errors). daemon-reload is likewise skipped when nothing was removed. macOS gets the same guarantee via the verified `launchdUnload` from the other thread.

internal/service/logs.go
jeremytondo: Fixed in 39193ae. The Linux branch now checks the unit file after the journalctl-exists check: on a never-supervised machine `atc server logs` errors with \"atc.server is not installed, so no supervised logs exist; foreground `atc server run` logs to stderr\" instead of printing an empty journal and succeeding. Covered in `TestRecoveryCommandsSurviveBrokenConfig` and smoke-tested live.

PR #247, opened 2026-08-25
ATC-266 | Restructure constructors to idiomatic Go: required deps as parameters, optionals defaulted

jeremytondo:
Implements [ATC-266](https://linear.app/elevenideas/issue/ATC-266/restructure-constructors-to-idiomatic-go-required-deps-as-parameters).

## What changed

- **`internal/server`**: the `Options` struct is gone. `NewHandler(verify, version, logger)` takes everything positionally — `verify` is the one true requirement (nil panics at construction with a clear message instead of on the first request), a nil `logger` defaults to `slog.DiscardHandler`, and `version` is just a string. The middleware helpers now take the specific values they use.
- **`internal/tailscale`**: `Supervisor` fields are unexported; construction goes through `NewSupervisor(executable, port, logger)` with the same nil-logger-to-discard default. The test-only `readyTimeout`/`waitDelay` knobs are unchanged.
- **Call sites and tests** updated; each package gains a `TestNilLoggerDefaultsToDiscard` exercising the paths that were latent nil-dereference panics (the version-skew log in the server, the retry-failure log in the supervisor).
- **"Chassis" retired** from the three production comments (`server.go` package doc, the removed `Options` doc, `internal/service` package doc). It survives only under `experiments/`, which stays untouched per AGENTS.md.

Nothing added to AGENTS.md — once the constructors follow the ecosystem norm, the code is the documentation.

## Verification

`mise check` passes: build, golangci-lint (0 issues), and all tests under the race detector.

### DIFF
--- a/cmd/atc/main.go
+++ b/cmd/atc/main.go
@@ -355,11 +355,7 @@ func serverRun(cmd *cobra.Command, _ []string) error {
 
 	version := versionString()
 	logger.Info("server starting", "version", version, "pid", os.Getpid())
-	handler := server.NewHandler(server.Options{
-		Version: version,
-		Verify:  store.Verify,
-		Logger:  logger,
-	})
+	handler := server.NewHandler(store.Verify, version, logger)
 
 	listener, err := net.Listen("tcp", net.JoinHostPort(cfg.Bind, strconv.Itoa(cfg.Port)))
 	if err != nil {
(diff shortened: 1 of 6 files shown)


jeremytondo (commented):
Reviewed against ATC-266 and the current server/tailscale contracts. I found no actionable issues: required dependencies now flow through constructors, nil loggers safely default to discard, middleware ordering is unchanged, and the Tailscale supervision lifecycle is preserved. Focused race-enabled tests pass for `internal/server` and `internal/tailscale`; CI is also green.

PR #248, opened 2026-08-26
ATC-261 | Shipping: GoReleaser release workflow, dev/production channels, install.sh, and atc upgrade

jeremytondo:
Implements [ATC-261](https://linear.app/elevenideas/issue/ATC-261) per the 2026-08-25 grill record.

## What's here

**Versioning** — new `internal/version` package: `Version` stamped via `-ldflags -X` by the release workflow only; the buildinfo `devel-<sha>`/`devel-<sha>-dirty` fallback stays for local builds, so only workflow builds ever claim a release-shaped version. `cmd/atc` now uses it everywhere `versionString()` was.

**GoReleaser** (`.goreleaser.yaml`) — CGO-free static builds for darwin/arm64, linux/amd64, linux/arm64; one tar.gz per target (binary + README) with **version-independent asset names** (`atc_<os>_<arch>.tar.gz`) so `releases/latest/download` URLs are static and the dev channel can clobber in place; `checksums.txt`; snapshot versioning `{{ incpatch .Version }}-dev.{{ .ShortCommit }}`; legacy and `dev` tags excluded from version computation (the `v0.0.x` tags stay visible so the first `minor` cut lands on v0.1.0).

**Release workflow** (`release.yml`) — `workflow_dispatch` with a single `bump` choice (`patch|minor|major|dev`). Production: computes the next version from the remote's tags, tags the dispatched ref's head, publishes an immutable release (stateless — safe from any machine). Dev: GoReleaser snapshot + `gh release upload --clobber` onto the rolling `dev` prerelease (created on first use, notes updated with the current build). `contents: write` scoped to the release job.

**install.sh** — first install only; OS/arch detection with named failures (Intel Macs get a real message, not a 404), checksum verification, `~/.local/bin` on both platforms, `ATC_INSTALL_DIR`/`ATC_VERSION` knobs, PATH hint. No dev knob by design.

**`atc upgrade`** — stateless channel: plain upgrade moves to latest production (short-circuits on version match; the dev→production walk-back is stated plainly), `--dev` always reinstalls. Tokenless discovery via the `releases/latest` redirect and the fixed `dev` tag. Swap: checksum verify → extract → stage beside the resolved `os.Executable()` → prove the staged binary runs (`version`) → atomic rename; unwritable dirs get a named diagnostic, never sudo. Post-swap: probe via the existing handshake; interactive runs are asked with the ATC-246 risk summary (default yes), **headless runs never restart** without `--restart`, `--no-restart` pre-answers the other way. A broken config.toml skips only the server check, never the swap.

**Conveniences** — mise: `goreleaser` tool, `install` task (local build into `~/.local/bin`), `release:patch|minor|major|dev` wrappers; README install/upgrade section.

## Verification

- `mise run check` green (build, golangci-lint, tests with race detector).
- `goreleaser check` + a full local `goreleaser release --snapshot --clean`: all three targets built, archives named `atc_<os>_<arch>.tar.gz` containing `atc` + README, snapshot version computed as `0.0.5-dev.15c7f1b`, and the extracted binary reports the stamped `v0.0.5-dev.15c7f1b`.
- Pure logic tested: asset naming (incl. named unsupported-platform diagnostics), redirect tag parsing, checksum verify, tar extraction, stage/promote with the unwritable-dir diagnostic, the restart policy matrix, prompt semantics, message rendering.

## Notes

- GoReleaser pinned at 2.17.1: mise's aqua backend fails GitHub artifact-attestation verification on 2.18.0 (recent releases are signed with GitHub's new `dotcom.releases.github.com` identity, which aqua doesn't expect yet).
- The archives ship `README.md` only — the repo has no LICENSE file to include.
- Until v0.1.0 exists, `releases/latest` still points at the archived product's v0.0.4; `atc upgrade` on this branch would fail there with the named "has this release been published?" diagnostic. Self-heals at the first production cut.

## After merge (the ticket's done-when)

1. `mise run release:minor` → cuts **v0.1.0**.
2. `mise run release:dev` → first rolling dev build.
3. `curl -fsSL .../install.sh | sh` on a machine, then one `atc upgrade` against a later release end-to-end.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,75 @@
+name: Release
+
+# One workflow, one input (ATC-261). patch/minor/major cut an immutable
+# production release: the next version is computed from the remote's
+# authoritative tags and stamped onto the head of the dispatched ref, so a
+# cut is safe from any machine regardless of checkout state. dev refreshes
+# the single rolling `dev` prerelease (GoReleaser snapshot + asset clobber)
+# and can be dispatched from any branch.
+
+on:
+  workflow_dispatch:
+    inputs:
+      bump:
+        description: "patch/minor/major cut a production release from this ref; dev refreshes the rolling dev build"
+        required: true
+        type: choice
+        options:
+          - patch
+          - minor
+          - major
+          - dev
+
+concurrency: release
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # GoReleaser derives changelogs and previous tags from history.
+          fetch-depth: 0
+      - uses: jdx/mise-action@3c2e0cf82a5b2e5249f0d3635a4d83d0ae861518 # v4.2.5
+      - name: Cut production release
+        if: inputs.bump != 'dev'
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+          BUMP: ${{ inputs.bump }}
+        run: |
+          git fetch --tags --force origin
+          latest=$(git tag --list 'v*' | { grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' || true; } | sort -V | tail -n1)
+          latest=${latest:-v0.0.0}
+          IFS=. read -r major minor patch <<<"${latest#v}"
+          case "$BUMP" in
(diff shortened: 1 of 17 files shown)


jeremytondo (commented):
Reviewed the full change against the surrounding implementation and the ATC-261 decided record. CI is green, and the focused local checks passed (`go test -race ./...`, `go vet ./...`, `goreleaser check`, and `sh -n install.sh`). I found six actionable release/reliability issues, noted inline.

PR #249, opened 2026-08-26
ATC-264 | Shared /v1 contract: internal/api wire structs, typed Problem, hand-written api.Client

jeremytondo:
Implements the decided record in [ATC-264](https://linear.app/elevenideas/issue/ATC-264/explore-and-decide-how-client-and-server-share-the-v1-api-contract): the Go structs are the single source of truth for the /v1 API shape, shared directly with every in-repo Go consumer.

## What changed

**New `internal/api` package** — the wire contract and its client, deliberately stdlib-only so a later promotion out of `internal/` for third-party Go consumers stays cheap:
- `Health` wire struct with Huma's validation/doc tags (inert to clients), plus the `Atc-Client-Version`/`Atc-Server-Version` header constants moved out of `internal/server` so client code never imports the server.
- Typed RFC 7807 `Problem` error (with `ErrorDetail` for Huma validation failures). It carries `ServerVersion` from the response header, so a tokenless probe reads the server's version off a 401. Callers branch with `errors.AsType` when they care and print the error when they don't.
- Hand-written `api.Client`: one shared request path sets the bearer token and client-version header and returns `*Problem` for any non-2xx — the skew handshake (ATC-247 §6) lives in exactly one place. A non-problem error body (proxy page, wrong process on the port) degrades to a synthesized problem with the right status. `Health(ctx)` is the first typed operation.

**`internal/server` rewired to wrap the shared structs** — `HealthOutput` wraps `api.Health` (named body types become reusable OpenAPI components), and the auth middleware's 401 body is now an encoded `api.Problem` instead of an ad-hoc map, making the one-error-contract claim structural.

**Probe migrated onto `api.Client`** — `probeOnce` in `internal/service` branches on `errors.AsType[*api.Problem]`, retiring the `service → server` import the ticket called out as the drift-shaped smell. `internal/server` is now imported only by `cmd/atc/main.go`.

## Verification

- `mise run check` green: build, golangci-lint, `go test -race ./...`.
- New tests cover the success path, header emission, the tokenless-401-carries-version contract, non-problem body degradation, and transport-error-vs-problem distinguishability; the server health test now decodes directly into `api.Health`.
- `go list -deps ./internal/api` confirms the package is stdlib-only.

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -1,0 +1,26 @@
+// Package api is the /v1 wire contract and its Go client (ATC-264). The
+// structs here are the single source of truth for the API shape: the
+// server's Huma handlers wrap them (the served OpenAPI document derives
+// from these same definitions), and every in-repo Go consumer imports
+// them, so a contract change that breaks a consumer is a compile error.
+// The validation/doc tags are Huma's; they are inert to clients.
+//
+// The package is deliberately stdlib-only, which keeps a later promotion
+// out of internal/ for third-party Go consumers cheap. Codegen exists only
+// across language boundaries: external consumers generate from the served
+// /openapi.json, never from a checked-in artifact and never in this repo.
+package api
+
+// Version headers ride both ways on every request/response — the entire
+// skew handshake (ATC-247 §6). They live here, not in the server, so
+// client-side code never imports server internals.
+const (
+	ClientVersionHeader = "Atc-Client-Version"
+	ServerVersionHeader = "Atc-Server-Version"
+)
+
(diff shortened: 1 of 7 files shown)


jeremytondo (commented):
The shared contract/client direction is solid and the full `mise run check` passes. I found four response-path issues to address before merging: successful calls currently drop the version handshake, malformed 2xx responses are mistaken for no HTTP response, problem bodies can override the authoritative HTTP status used for branching, and the unbounded deferred drain defeats the body-size cap.

PR #255, opened 2026-08-28
ATC-273 | cli: extract client construction, project resolution, and attach into internal/cli

jeremytondo:
Client-side behavior had accumulated in `cmd/atc` (the ATC-256 walk-up resolution, the ATC-271 attach machinery, and the API-client connection policy), leaving real functionality unreachable outside `package main`. This moves it into a new `internal/cli` package and reduces `cmd/atc` to cobra wiring, flag parsing, and output formatting.

Linear: [ATC-273](https://linear.app/elevenideas/issue/ATC-273/extract-client-side-cli-logic-from-cmdatc-into-internalcli)

## What moved

- **`internal/cli/client.go`** — `NewClient`: the shared API client and its connection policy (ATC_SERVER / ATC_TOKEN over the settled local config port and token file, plus the one-line version-skew warning). Was `newAPIClient` in `cmd/atc/terminal.go`.
- **`internal/cli/project.go`** — `ResolveProjectID` (canonical-cwd walk-up, nearest ancestor wins, TTY create-offer prompt on no match), `DefaultProjectDir`, and `ProjectRootFor` (git-authoritative toplevel). Was in `cmd/atc/project.go`.
- **`internal/cli/attach.go`** — `AttachPreflight` (TTY and loopback checks), `AttachSession` (running-status check, socket validation, the `syscall.Exec` handover), and the `isLoopback` helper. Was in `cmd/atc/terminal.go`.

The `stdinIsTTY` / `stdioIsTerminal` test seams moved with their consumers as `cli.StdinIsTTY` / `cli.StdioIsTerminal`.

## What stayed

- `cmd/atc` keeps the command tree, flag definitions and parsing, help text, tabwriter printers, and the `runWithClient` wrapper (now over `cli.NewClient`).
- The server composition root stays in `main.go` — main is the conventional home for wiring the store, adapter, services, and listener together, and nothing else needs to reuse it.
- The CLI tests stay in `cmd/atc` unchanged apart from the seam references: they exercise the full command surface through `run()`, which is the contract they cover.

No behavior changes; this is a pure move. Motivation beyond tidiness: the TUI (ATC-258) needs exactly this client-side behavior and could not reach it inside `package main`.

`mise run check` passes (build, golangci-lint, sqlc diff, `go test -race ./...`).

## Summary by CodeRabbit

- **New Features**
  - Added streamlined terminal attachment support for interactive sessions.
  - Added automatic project-directory selection when no path is specified.
  - Added shared CLI setup for server connection and authentication.

- **Bug Fixes**
  - Terminal attachment now validates interactive use and permits only local server connections.
  - Added a single warning when client and server versions differ, including during concurrent requests.
  - Improved project detection and Git repository handling.

### DIFF
--- a/internal/cli/attach.go
+++ b/internal/cli/attach.go
@@ -13,9 +13,11 @@ import (
 	"github.com/jeremytondo/atc/internal/zmx"
 )
 
-// StdioIsTerminal reports whether stdin and stdout are both real TTYs.
-// It is a variable so tests, which never have a TTY, can force it.
-var StdioIsTerminal = func() bool {
+// StdioIsTerminal reports whether this process's stdin and stdout are both
+// real TTYs. Callers pass the result (or their own knowledge) into
+// AttachPreflight, so the check stays an explicit capability rather than
+// package state.
+func StdioIsTerminal() bool {
 	stdin, err := os.Stdin.Stat()
 	if err != nil || stdin.Mode()&os.ModeCharDevice == 0 {
 		return false
@@ -25,10 +27,10 @@ var StdioIsTerminal = func() bool {
 }
 
 // AttachPreflight refuses attach attempts that can never succeed: a
-// non-interactive stdio pair, or a non-loopback server (the session socket
-// lives on the server's machine).
-func AttachPreflight(baseURL string) error {
-	if !StdioIsTerminal() {
+// non-interactive stdio pair (stdioIsTerminal, supplied by the caller), or
+// a non-loopback server (the session socket lives on the server's machine).
+func AttachPreflight(baseURL string, stdioIsTerminal bool) error {
+	if !stdioIsTerminal {
 		return fmt.Errorf("attaching requires an interactive terminal (stdin and stdout must be TTYs)")
 	}
 	if !isLoopback(baseURL) {
--- a/internal/cli/attach.go
+++ b/internal/cli/attach.go
@@ -1,0 +1,79 @@
+package cli
+
+import (
+	"fmt"
+	"net"
+	"net/url"
+	"os"
+	"path/filepath"
+	"syscall"
+
+	"github.com/jeremytondo/atc/internal/api"
+	"github.com/jeremytondo/atc/internal/paths"
+	"github.com/jeremytondo/atc/internal/zmx"
+)
+
+// StdioIsTerminal reports whether stdin and stdout are both real TTYs.
+// It is a variable so tests, which never have a TTY, can force it.
+var StdioIsTerminal = func() bool {
+	stdin, err := os.Stdin.Stat()
+	if err != nil || stdin.Mode()&os.ModeCharDevice == 0 {
+		return false
+	}
+	stdout, err := os.Stdout.Stat()
+	return err == nil && stdout.Mode()&os.ModeCharDevice != 0
+}
+
+// AttachPreflight refuses attach attempts that can never succeed: a
+// non-interactive stdio pair, or a non-loopback server (the session socket
+// lives on the server's machine).
+func AttachPreflight(baseURL string) error {
+	if !StdioIsTerminal() {
+		return fmt.Errorf("attaching requires an interactive terminal (stdin and stdout must be TTYs)")
+	}
+	if !isLoopback(baseURL) {
+		return fmt.Errorf("atc terminal attach is local-only (the session socket lives on the server's machine); this client targets %s", baseURL)
+	}
+	return nil
+}
+
+// AttachSession hands this process's TTY over to the terminal's running
+// session. On success it never returns: the process is replaced by zmx
+// until detach.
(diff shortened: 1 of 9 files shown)


internal/cli/attach.go
```diff
@@ -0,0 +1,79 @@
...
+	"syscall"
+
+	"github.com/jeremytondo/atc/internal/api"
+	"github.com/jeremytondo/atc/internal/paths"
+	"github.com/jeremytondo/atc/internal/zmx"
+)
+
+// StdioIsTerminal reports whether stdin and stdout are both real TTYs.
+// It is a variable so tests, which never have a TTY, can force it.
+var StdioIsTerminal = func() bool {
```
jeremytondo: `StdioIsTerminal` (and the matching `StdinIsTTY`) is now exported mutable package state solely so `cmd/atc` tests can replace it. That lets any client silently change every other client’s behavior, couples tests through a process-global variable, and prevents safe parallel tests—the opposite of a reusable layer for both Cobra and the TUI. Keep the detection dependency explicit at the operation/client boundary (or make it an internal dependency of a constructed runtime) so callers provide their own stdio/capabilities and tests do not mutate global state.

internal/cli/project.go
```diff
@@ -0,0 +1,101 @@
...
+// toplevel when inside a repository, the current directory otherwise.
+func DefaultProjectDir() (string, error) {
+	canonical, err := paths.CanonicalDir(".")
+	if err != nil {
+		return "", err
+	}
+	return ProjectRootFor(canonical), nil
+}
+
+// ProjectRootFor asks git for the toplevel of the repository containing
```
jeremytondo: `ProjectRootFor` is exported only because `cmd/atc/project_test.go` calls it for a test-environment skip; every production caller is in this package. This expands the new shared package API with an implementation detail and invites the TUI to couple to its git subprocess behavior. Keep it unexported and test the public `DefaultProjectDir`/resolution behavior instead (or move this focused test beside the package).

jeremytondo (commented):
Simplicity review: the extraction is otherwise a faithful, focused move. I found two API-boundary issues that will make the intended shared client layer harder to reuse and test.

internal/cli/client.go
```diff
@@ -0,0 +1,58 @@
...
+		tokenPath, err := paths.AuthTokenFile()
+		if err != nil {
+			return nil, "", err
+		}
+		if token, err = (&authtoken.Store{Path: tokenPath}).Ensure(); err != nil {
+			return nil, "", err
+		}
+	}
+	clientVersion := version.String()
+	warned := false
```
jeremytondo: **Correctness:** Concurrent use of the extracted client races on `warned`. `api.Client.send` invokes `onServerVersion` on the goroutine issuing each request, and the underlying `http.Client` supports concurrent calls. That is especially relevant to this extraction’s intended TUI consumer: two in-flight responses can read/write this bool concurrently, producing a `-race` failure and potentially duplicate warning output. Please guard the one-time warning with `sync.Once` (or equivalent) and cover concurrent use at the reusable `cli.NewClient` boundary.

jeremytondo (commented):
Correctness review: 1 actionable issue found. The serial Cobra flows preserve the moved behavior, and `mise run check` passes (build, lint, sqlc diff, and race-enabled tests). However, the newly reusable client introduces unsynchronized per-client warning state, so concurrent consumers can race; see the inline comment. I recommend fixing that before this boundary is consumed by the TUI.

internal/cli/client.go
jeremytondo: Fixed in 0110b44c. The one-time warning is now guarded by a `sync.Once` inside the `onServerVersion` closure, and a new `internal/cli/client_test.go` covers the reusable boundary: `TestNewClientVersionSkewWarnsOnceUnderConcurrency` fires 8 concurrent requests against a skewed-version test server through `cli.NewClient` and asserts exactly one warning line. Passes under `go test -race`.

internal/cli/attach.go
jeremytondo: Fixed in 0110b44c. `StdioIsTerminal` and `StdinIsTTY` are now pure exported functions — no mutable package state in `internal/cli`. The operations take the capability explicitly (`AttachPreflight(baseURL, stdioIsTerminal bool)`, `ResolveProjectID(..., stdinIsTTY bool)`), so callers provide their own stdio knowledge. The test seams moved back into `package main` as unexported vars defaulting to the cli detection functions, restoring the pre-extraction pattern without exporting it.

internal/cli/project.go
jeremytondo: Fixed in 0110b44c. `projectRootFor` is unexported again; every production caller is in-package. The `cmd/atc` test skip guard now probes git directly (`git -C <dir> rev-parse --show-toplevel`) — the same authority the CLI consults — instead of reaching into the package.

